PR #596, opened 2020-06-03
Link dependencies dynamically

### LINKED ISSUE #595: Link all direct dependencies dynamically
We are producing a dynamic library, so we should also link dynamically to other libraries. Besides being a good idea in general, this also helps to avoid licensing issues, because then we're not including third-party code in the binaries we're distributing.

See also #594.

kyllingstad:
This fixes #595.

I'm only setting `shared=True` for our direct dependencies, since those are the ones we link against, so I also reordered the `requires` list to make it obvious which ones they are.


kyllingstad:
This fails on Windows because the `fmilibrary` package isn't available in Artifactory with `shared=true`. Not sure what the best way to fix this is. Should I add `-b missing` to the workflow?

kyllingstad:
Nevermind, I see that it's supposed to be built by the win-conan workflows, but they're failing for some other reason. I'll fix that.

PR #600, opened 2020-06-22
Use more conventional Conan channel names

### LINKED ISSUE #591: Use more conventional Conan channel names
The packages we push to Artifactory should use more conventional channel names. Having to use package references like `libcosim/<version>@osp/<version>` is just going to trip people up.

I suggest the following:
* `osp/stable` for packages built from release tags
* `osp/testing` for packages built from `master`
* `osp/testing-<branch>` for packages built from other branches

kyllingstad:
I recently did this for some internal software, so it felt like a good time to do it here too.

If this passes review, I'll make corresponding PRs to other repos too before merging it, so we can merge all at once.

Fixes #591.

### DIFF
--- a/.github/workflows/ci-conan.yml
+++ b/.github/workflows/ci-conan.yml
@@ -36,18 +36,13 @@ jobs:
         run: conan remote add osp https://osp-conan.azurewebsites.net/artifactory/api/conan/public --force
       - name: Conan create
         run: |
-          BRANCH=$(git rev-parse --abbrev-ref HEAD)
-          TAGS=$(git tag --points-at HEAD)
-          VERSION=$(<version.txt)
-          if [ "$BRANCH" = "master" ]; then
-            if { echo "$TAGS" | grep -q "^v$VERSION\$"; }; then
-              CHANNEL="stable"
-            else
-              CHANNEL="testing"
-            fi
+          REFNAME="${GITHUB_REF#refs/*/}"
+          VERSION="v$(<version.txt)"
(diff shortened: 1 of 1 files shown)


.github/workflows/ci-conan.yml
```diff
@@ -36,9 +36,19 @@ jobs:
         run: conan remote add osp https://osp-conan.azurewebsites.net/artifactory/api/conan/public --force
       - name: Conan create
         run: |
-          BRANCH=${GITHUB_REF#refs/*/}
-          SHORT_BRANCH=${BRANCH:0:50}
-          CHANNEL=${SHORT_BRANCH//\//_}
+          BRANCH=$(git rev-parse --abbrev-ref HEAD)
```
hplatou, line 39: Running this command on a locally checked out tag(v0.4.0) only returns 'HEAD'. Running when checked out the master branch correctly returns 'master'

kyllingstad: Wow, good catch! I was under the mistaken impression that release builds were also done off `master`, and that what distinguishes them from other builds is whether there's a version tag present on the commit too.

I've pushed a fix that reverts to using `GITHUB_REF` again.

kyllingstad:
I've updated `libcosimc` and `cosim-cli` now. `cosim-demo-app` and `libcosim-consumer` are still using v0.7.0 (which retains the old channel names), so we just have to remember to change the channel too when they start using a newer version.

Once the other two PRs are approved, I'll merge all three in sequence, ensuring they all go green before merge.

PR #605, opened 2020-06-24
Real time timer improvements

### LINKED ISSUE #298: Measured real time factor is wrong
Under certain circumstances, the real time factor measured by `cse::real_time_timer` is wrong.

Reproduction:
- Load cse-demos/dp-ship with cse demo application.
- Override value of "dpsi_ref" input of "DP Controller" with some nonzero value.

Result:
- Displayed real time factor in GUI is ~2.xxx, but simulation time progresses at a much slower rate.

Cause:
- One of the models experiences a "hickup" every second or so, spending a lot of time on one step. Real time factor is measured every five steps, thus we are not catching the "slow" step.

eidekrist:
This is an attempt to fix #298, measured real time factor is wrong.

Two new features are introduced:
- Number of steps used in the rolling average RTF calculation is now configurable.
- Total average real time factor (since start of simulation or real time factor toggling) is calculated and exposed to the outside.

I was not happy with the growing amount of timer-related methods delegated through the `execution` class, so I introduced two new data classes:
- `real_time_config` for toggling real time simulation, setting RTF target etc.
- `real_time_metrics` for monitoring of real time performance.

Please let me know if this was a bad idea and/or implementation!

Will create separate PR's for client code if this gets approved.

### DIFF
--- a/src/cosim/timer.cpp
+++ b/src/cosim/timer.cpp
@@ -5,8 +5,6 @@
  */
 #include "cosim/timer.hpp"
 
-#include "cosim/error.hpp"
-
 #include <chrono>
 #include <thread>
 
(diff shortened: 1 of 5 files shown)


src/cosim/timer.cpp
```diff
@@ -18,11 +16,55 @@ constexpr std::chrono::microseconds MIN_SLEEP(100);
 namespace cosim
 {
 
+void real_time_config::set_real_time_simulation(bool enable)
+{
+    if (realTimeSimulation_ != enable) {
+        configChanged_.store(true);
```
kyllingstad, line 22: Even though the individual variables are atomic, you could end up in a situation like the following, where thread A wants to change the configuration and thread B wants to read it:

1. Thread A sets `configChanged_` to `true`.
2. Thread B sees that `configChanged_` is `true`, reads `realTimeSimulation_`.
3. Thread A sets `realTimeSimulation_` to new value.
4. Thread B sets `configChanged_` to `false`.

In other words, B gets the old value even though it deduced that the configuration had changed, and at the end of it all, the configuration is marked ad "not changed" even though it is.

I'm not 100% sure whether that's a problem here, but maybe you can tell?

eidekrist: You are right! The consequences shouldn't be *terrible*, but it means I overthought/engineered this part and can probably simplify it. Will update!

eidekrist:
I have pushed an update with some changes.

`real_time_config` is now a plain struct, like `real_time_metrics`. Changes are detected by computing the hash of the local `config_` object and comparing it to a stored value. With regards to concurrency issues, a value updated by another thread could potentially be used for one step before the change is detected, which is not critical. 

Functionality wise, setting `real_time_factor_target` to a nonpositive value no longer throws, but will effectively disable real-time synchronization.

src/cosim/timer.cpp
```diff
@@ -31,83 +30,59 @@ class real_time_timer::impl
...
-        lastSimulationTime_ = currentTime;
-        if (realTimeSimulation_) {
-            const auto elapsed = current - startTime_;
-            const auto expectedSimulationTime = (currentTime - simulationStartTime_) / realTimeFactorTarget_.load();
+        const auto newHash = std::hash<real_time_config>()(*config_);
+        if (newHash != configHashValue_) {
+            start(currentTime);
+            configHashValue_ = newHash;
+        }
+        if (config_->real_time_simulation && config_->real_time_factor_target.load() > 0.0) {
```
kyllingstad, line 42: You call `config_->real_time_factor_target.load()` twice here, first in line 42 and then in line 44. But if the `real_time_config` struct is designed to be accessed by multiple threads simultaneously, then it may change between these two loads. It could therefore be `> 0.0` at first, so that the `if` block is entered, but then be `== 0.0` on the second load, triggering a divide-by-zero error.

A simple solution would be to load it to some local variable and use that instead.

eidekrist: Dang! Will fix.

PR #610, opened 2020-09-03
Update thrift (0.13.0)

markaren:
Updates thrift to 0.13.0

Even with the change, libcosim is still compatible with fmu-proxy 0.61, 0.6.2


hplatou (approved):
Looks good to me.

PR #615, opened 2020-09-21
Upgraded libzip dependency to new version available in ConanCenter

hplatou:
Upgraded the libzip conan dependency to 1.7.3, that is now available through ConanCenter. The previoius version 1.5.2 available through bincrafters is now marked obsolete. The transitive dependencies of the new version is compatible with the rest of our dependencies and I could remove the dependency conflict resolution as well. This also resolves some runtime issues when linking an executable with both libcosim and Poco(with transitive openssl) from ConanCenter.

### DIFF
--- a/conanfile.py
+++ b/conanfile.py
@@ -19,13 +19,9 @@ class LibcosimConan(ConanFile):
         "boost/1.71.0",
         "fmilibrary/2.0.3",
         "ms-gsl/2.1.0",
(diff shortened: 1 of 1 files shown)


kyllingstad (approved):
Awesome that we get rid of the conflict-resolving explicit dependencies!

markaren (approved):

PR #617, opened 2020-10-06
Add an introductory page to the API documentation

### LINKED ISSUE #616: Add start page to API documentation
Every time I need to look up something in the libcosim API documentation, I have to go via the namespace/class/file lists generated by Doxygen. All of these have become rather large, and sifting through them takes time. It would be great to have a start page that allows one to find things more quickly.

While our API consists of a large number of classes and functions, some of them are more central than others. Examples include `execution`, `system_structure`, `slave`, `observer` and so on. These could serve as starting points for exploring the API, as their documentation links to that of other, less central entities that mainly serve as support structure.

This would also make the documentation more accessible to newcomers, who have *no* idea of where to start.

kyllingstad:
The contents of `docs/main_page.md` will be the first thing the user sees upon browsing the API documentation. That is, it will fill the current void below the menu [here](https://open-simulation-platform.github.io/libcosim/doxygen/libcosim/0.7.0/index.html) (substituting "Introduction" in place of the "libcosim Documentation" heading). All symbols link to the appropriate place in the API documentation, by Doxygen magic.

This fixes #616.

### DIFF
--- a/docs/main_page.md
+++ b/docs/main_page.md
@@ -1,0 +1,166 @@
+\mainpage Introduction
+
+The [libcosim](https://open-simulation-platform.github.io/libcosim) API can
+roughly be divided into two parts, corresponding to the main phases of a
+co-simulation. These are:
+
+  * The _setup phase_, which is everything that happens before you actually
+    run a simulation, such as reading configuration files, loading FMUs,
+    setting up the system for simulation, and so on.
+  * The _execution phase_, where we run a simulation, solving the model
+    equations, logging and/or displaying output, et cetera.
+
+These phases are not distinguished in any particular way in the API. Most
+symbols simply live in the top-level `#cosim` namespace, and there is no
+specific naming convention to set them apart. Nor are they entirely distinct
+in functionality, as several classes and functions are used in both phases.
+Nevertheless, the distinction between "setup" and "execution" provides a good
+framework for learning and understanding the libcosim API.
+
+## Setting up a simulation
+
+The central class in the setup phase is `cosim::system_structure`. This
+contains a representation of the modelled system, consisting of the entities
+(simulators and functions) in it and the connections between them. It requires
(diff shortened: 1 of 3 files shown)


docs/main_page.md
```diff
@@ -0,0 +1,166 @@
...
+specific naming convention to set them apart. Nor are they entirely distinct
+in functionality, as several classes and functions are used in both phases.
+Nevertheless, the distinction between "setup" and "execution" provides a good
+framework for learning and understanding the libcosim API.
+
+## Setting up a simulation
+
+The central class in the setup phase is `cosim::system_structure`. This
+contains a representation of the modelled system, consisting of the entities
+(simulators and functions) in it and the connections between them. It requires
```
hplatou, line 24: Don't remember if we have concluded on this before. Are we going to refer to this entity as a `simulator` or a `model`? In our code system_structure is using `cosim:model`

hplatou: Btw. This doc is very nice!

markaren:
Content looks good. I can't build and test the doc myself, so I can't vouch for that.

docs/main_page.md
kyllingstad: Thanks! :)

In my mind, the model/simulator distinction is similar to the class/object distinction in OOP, or, more directly, to the FMU/instance distinction in FMI. And that's the way the API is designed too: `cosim::model` is a factory class (which typically represents an FMU) which has a method to create simulator instances.

I wasn't sure about the wording in this part of the text myself. I landed on "simulator" because, while `system_structure` uses the `model` class, it uses it to represent what _will be_ a simulator in the instantiated system. Similarly, there are no actual "connections" here, because there are no instances to connect, but I still used "connections" to refer to the representation of _future_ connections.

hplatou: 👍

hplatou (approved):

Issue #608, opened 2020-08-03
Please log in to "osp" to perform this action


andre2007:
As far as I understand the error log, you cannot use conan without registration? Could you add to the README where I have to register and maybe the missing conan login commands?

``` dockerfile
FROM debian:buster

RUN apt-get update && apt-get upgrade -y \
    && apt-get install -y curl unzip build-essential gcc cmake python3-pip
    
RUN pip3 install conan

RUN curl -L -o libcosim.zip https://github.com/open-simulation-platform/libcosim/archive/v0.7.0.zip \
    && unzip libcosim.zip -d / \
    && cd /libcosim-0.7.0 \
    && mkdir build \
    && cd build \
    && conan remote add osp https://osp.jfrog.io/artifactory/api/conan/conan-public \
    && conan install .. -s build_type=Release --build=missing -o fmuproxy=True --settings=compiler.libcxx=libstdc++11
```

> fmilibrary/2.0.3: Not found in local cache, looking in remotes...
fmilibrary/2.0.3: Trying with 'conan-center'...
fmilibrary/2.0.3: Trying with 'osp'...
Please log in to "osp" to perform this action. Execute "conan user" command.
Remote 'osp' username: ERROR: Failed requirement 'fmilibrary/2.0.3' from 'conanfile.py (libcosim/0.7.0)'
ERROR: EOF when reading a line. [Remote: osp]

markaren:
Looks like the newly updated osp remote needs credentials (which it should not). @hplatou

hplatou:
We are in the middle of a process to move to a new Conan repository(you are using the new correct url) and the permissions were not set correctly. I have updated now so please try again.

andre2007:
You might have to adapt the README, as there are some more flags needed:

ERROR: Failed requirement 'fmilibrary/2.0.3' from 'conanfile.py (libcosim/0.7.0)'
ERROR: The remote at 'https://osp.jfrog.io/artifactory/api/conan/conan-public' only works with revisions enabled. Set CONAN_REVISIONS_ENABLED=1 or set 'general.revisions_enabled = 1' at the 'conan.conf'. [Remote: osp]

andre2007:
Also just setting the revisions enabled flag does not work:
`conan install .. -s build_type=Release --build=missing -o fmuproxy=True --settings=compiler.libcxx=libstdc++11 --settings=general.revisions_enabled=1`

ERROR: general not defined for general.revisions_enabled
Please define general value first too

[...]

PR #622, opened 2020-11-25
Update README.md

ljamt:
Adding 
- Links to applications using libcosim
- How to use
- CONAN_REVISIONS_ENABLED 
- conan/local instead of conan/public

Closing #608

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,7 +24,7 @@ following these steps:
 * Add the OSP Conan repository as a remote: 
 
        conan remote add osp https://osp.jfrog.io/artifactory/api/conan/local
-* As our repository has enabled revision support you must set the envorinmental variable `CONAN_REVISIONS_ENABLED=1`
+* Package revisions must be enabled. See [How to activate the revisions].
 * Include libcosim as a requirement in your conanfile 
 * Run `conan install` to aquire the libcosim package
 
@@ -50,9 +50,7 @@ First, add the OSP Conan repository as a remote:
 
     conan remote add osp https://osp.jfrog.io/artifactory/api/conan/local
 
-The Conan repository has enabled revision support meaning that you must set 
-the environmental variable `CONAN_REVISIONS_ENABLED=1` for your client to be
-compatible.
+Package revisions must be enabled. See [How to activate the revisions].
 
 As we will build the library using the *debug* configuration in this guide (as
 opposed to *release*), we must use the Conan setting `build_type=Debug`.  For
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -8,6 +8,27 @@ This repository contains the OSP C++ library for co-simulations.
...
+To use libcosim in your application you either build the library as described 
+in the section below or you can use conan. As libcosim is made available as a 
+conan package on https://osp.jfrog.io, you can include it in your application
+following these steps:
+
+* Install [Conan] >= 1.7
+* Add the OSP Conan repository as a remote: 
+
+       conan remote add osp https://osp.jfrog.io/artifactory/api/conan/local
+* As our repository has enabled revision support you must set the envorinmental variable `CONAN_REVISIONS_ENABLED=1`
```
kyllingstad, line 27:
```suggestion
* As our repository has enabled revision support you must set the environmental variable `CONAN_REVISIONS_ENABLED=1`
```

README.md
```diff
@@ -27,7 +48,11 @@ of course you can always build and install dependencies manually.
 
 First, add the OSP Conan repository as a remote:
 
-    conan remote add osp https://osp.jfrog.io/artifactory/api/conan/conan-public
+    conan remote add osp https://osp.jfrog.io/artifactory/api/conan/local
+
+The Conan repository has enabled revision support meaning that you must set 
+the envorinmental variable `CONAN_REVISIONS_ENABLED=1` for you client to be
```
kyllingstad, line 54:
```suggestion
the environmental variable `CONAN_REVISIONS_ENABLED=1` for your client to be
```

README.md
kyllingstad: It's also possible to enable revisions via a conan configuration file setting. So instead of being explicit (here and further down) about how, maybe we should just say that revisions must be enabled and link to the relevant part of the Conan docs:
https://docs.conan.io/en/latest/versioning/revisions.html?highlight=revisions#how-to-activate-the-revisions

kyllingstad (changes requested):

README.md
ljamt: Yes, my first attempt included using conan.conf, but it got too detailed. Including a link tothe Conan doc, was a good idea.

markaren:
An improvement, but how about splitting the BUILDING and INSTALLATION part out of the README?
That it, README, BUILDING and INSTALLING are three separate pages. README would link to building and installing like so:

image

kyllingstad (approved):
Great! Good to get this info in place, I know some people have got stuck on it.

ljamt:
> An improvement, but how about splitting the BUILDING and INSTALLATION part out of the README?

With the aim of slimming down the README, and provide better overview? I can split it into more files, but I'm good with it either way.

markaren:
> With the aim of slimming down the README, and provide better overview?

Precisely. However, I'm good with this change and the split could potentially be another PR.

markaren (approved):

PR #627, opened 2020-12-18
Feature/626 compile static libcosim

hplatou:
Added Conan option for building libcosim as static library. Also extended build matrix to include shared [True, False]. Please do some basic testing by using libcosim as static (libcosim:shared=False) dependency in consumers.


hplatou:
Needs fix for merge conflict.

.github/workflows/ci-cmake.yml
```diff
@@ -71,6 +73,9 @@ jobs:
       matrix:
         os: [windows-2016, windows-2019]
         build_type: [Debug, Release]
+        shared: ["True", "False"]
```
kyllingstad, line 76: It says `shared: ["True", "False"]` here while it says `shared_libs: [ON, OFF]` on line 25. This is probably intentional, but I just wanted to point out the inconsistency in case it was accidental.

kyllingstad (approved):
Looks good!

kyllingstad:
Btw, I tried to test this locally by building cosim CLI with the `libcosim:shared=False` option. I can confirm that the downloaded `libcosim` package only contains a static library and that it seems to depend in turn on static libraries, so this PR seems to work as intended.

However, I couldn't build cosim CLI, because *its* build system doesn't support static linking of Boost. So this would need to be fixed too.

src/CMakeLists.txt
```diff
@@ -190,8 +194,9 @@ if(CMAKE_CXX_COMPILER_ID STREQUAL "GNU")
     target_link_libraries(cosim INTERFACE "atomic" "stdc++fs")
 endif()
 if(WIN32)
+    target_link_libraries(cosim PUBLIC "Bcrypt")
     if(BUILD_SHARED_LIBS)
-        target_link_libraries(cosim PUBLIC "Bcrypt" "Shlwapi") # Windows API DLLs
+        target_link_libraries(cosim PUBLIC "Shlwapi") # Windows API DLLs
     else()
         set_target_properties(cosim PROPERTIES OUTPUT_NAME "libcosim")
```
ljamt, line 201: Shouldn't we have `OUTPUT_NAME "cosim"`?

.github/workflows/ci-cmake.yml
ljamt: Sorry, I don't know.

src/CMakeLists.txt
kyllingstad: Nope. Windows DLLs come with a companion `.lib` called the "import library" which is what you actually link into your application. The point here is to have different names for the import library and the static library so both can be installed alongside each other. The former is called `cosim.lib` to match the name of `cosim.dll`, while the latter is called `libcosim.lib`, borrowing the *NIX convention. (Boost does it the same way.)

Import libraries are not a thing on Linux, by the way. There, the linker is able to get the required information directly from the `.so` file.

kyllingstad (approved):

PR #631, opened 2021-03-29
Refactor/fmu

markaren:
Refactor to simplify fmilib interaction.

The old code was parsing the XML n+1 times for each fmu, where n is the number of slaves.
Also removed some needless calls to free et.al., as these are handled by the destructors anyway. 

Moved `createdllfmu` to the fmu itself.


markaren:
@kyllingstad pointed out that this has been implemented like this due to a deficiency in fmilib. I ran some tests to confirm it.

PR #632, opened 2021-03-30
Conan conflict resolution

markaren:
I guess this conflict appeared because we are not using revisions. Perhaps we should?
This update requires an up-to-date conan version.


kyllingstad (approved):

markaren:
Merging this so that others PR's does not fail. Other alternatives can be tested in subsequent PRs

PR #634, opened 2021-04-06
Add notices about workaround in FMU loader code

kyllingstad:
As PR #631 showed, it is not obvious why the `fmu` and `slave_instance` classes are implemented the way they are. (Even I, who wrote the code in the first place, had to dig around in the FMI Library sources to remember why...) Here I've added brief explanations for future reference.


ljamt (approved):

ljamt:
@kyllingstad, any problems with squash and merge this one?

kyllingstad:
Nope, I just forgot about it. :)

PR #635, opened 2021-04-13
Remove gcc8 from the workflow

markaren:
The ubuntu image no longer ships with gcc8.
https://github.com/actions/virtual-environments/commit/475453f2121f3e575d2f23c6ba4aa883f08de702

This PR removes it.

### DIFF
--- a/.github/workflows/ci-conan.yml
+++ b/.github/workflows/ci-conan.yml
@@ -22,7 +22,7 @@ jobs:
       matrix:
         os: [ubuntu-18.04]
         build_type: [Debug, Release]
-        compiler_version: [7, 8, 9]
+        compiler_version: [7, 9]
         compiler_libcxx: [libstdc++11]
         option_fmuproxy: ['fmuproxy=True', 'fmuproxy=False']
 


kyllingstad:
I guess this change means we'll lose the pre-built Conan packages for GCC 8 in Artifactory?

We (as in, my little group in SINTEF) are using Debian Stable as our main development platform, and the current compiler there is GCC 8. So this change will be a hassle for us. I don't know how representative we are, though.

markaren:
An alternative would be to manually add it back to the image?
But there has to be a reason they choose to remove it right?

markaren:
> The motivation for the changes
We are going to deprecate these versions of the listed tools due to lack of free space on Ubuntu images

> Mitigation ways
If you need to continue using these versions in your build, please consider the following ways to install it in runtime:

```
sudo apt-get update
sudo apt-get install g++-<version> -y

# For example, use this command to install gcc 8
sudo apt-get install g++-8 -y
```

ljamt:
I suggest we install gcc-8 and g++-8 in the workflows the way you describe @markaren

markaren:
Aight, lets review and merge this

kyllingstad (approved):

ljamt (approved):

PR #637, opened 2021-04-17
Feature/std filesystem

markaren:
This PR replaces usage of boost::filesystem with std::filesystem.
In order to support gcc7, the header `fs_portability` is used, which falls back to std::experimental::filesystem.
Thus, usage of `std::filesystem` should be done using the typealias `cosim::filesystem`

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -101,7 +101,7 @@ if(LIBCOSIM_USING_CONAN)
     endif()
 endif()
 
-set(Boost_components date_time fiber filesystem log)
+set(Boost_components date_time fiber log)
 if (LIBCOSIM_BUILD_TESTS)
     list(APPEND Boost_components timer unit_test_framework)
 endif()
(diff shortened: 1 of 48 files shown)


markaren:
The PR fails on file locking on windows.. Don't know how to fix..

markaren:
I hope this PR is welcome.  No need to use boost when we got the std available. The only caveat is that we need the fallback header in order to support gcc 6 and 7. We also need to fix the file locking unit test. @kyllingstad need your expertise on that.

kyllingstad:
Hopefully, this does the trick. The file locking mechanism has an internal cache of mutexes associated with different files, and each time one tries to create a new one, it checks whether a mutex exists for the file already. Since different paths (relative vs. absolute, links, etc.) can refer to the same file, it uses `filesystem::equivalent()` to perform this check. But this requires that the file exists, which we previously _didn't_ check.

kyllingstad (approved):
Approving @markaren's changes here. I suppose he or @ljamt should have a look at mine.

ljamt:
> Approving @markaren's changes here. I suppose he or @ljamt should have a look at mine.

Looks good to me 👍

ljamt (approved):

PR #638, opened 2021-04-23
Add default destructors to function objects

### LINKED ISSUE #621: Libcosim fails to compile with Visual Studio 2019
Visual Studio 2019 report warnings during compilation of libcosim:

`warning C5205: delete of an abstract class 'cosim::function' that has a non-virtual destructor results in undefined behavior`

This is reported for every class implementing the `function` interface and by default warnings will fail the build.

markaren:
Add default destructors so that  the code compiles on Visual Studio 2019

Closes #621

### DIFF
--- a/include/cosim/function/function.hpp
+++ b/include/cosim/function/function.hpp
@@ -169,6 +169,8 @@ public:
 
     /// Performs the function calculations.
     virtual void calculate() = 0;
+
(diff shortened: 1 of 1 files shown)


markaren:
@kyllingstad, time to merge this?

kyllingstad (approved):
Absolutely! :)

PR #633, opened 2021-04-06
Feature/proxy-fmu

markaren:
Adds proxy-fmu support to libcosim. proxy-fmu is implemented in C++ and supports automatic spawning of new processes. Thus it replaces the current usage of the JVM based fmu-proxy solution.

Closes #628

PR #644, opened 2021-08-17
Clean up Conan CI build environment using Docker and run tests in Conan CI builds

kyllingstad:
This change was motivated by some test failures in PR #633 which are almost surely unrelated to the PR itself. The probable diagnosis is simply that the build environment provided by GitHub's workflow runners is so messy that some unwanted library inadvertently gets picked up from *outside* the Conan package ecosystem. (We add to the mess by installing an additional GCC version, which happens to be the version for which the tests fail...)

Long story short, here I've set up a cleaner build environment using Docker. The building still takes place on GitHub's runners, but is now isolated from the messy host environment inside a Docker container. The base environments for the different compilers are [the ones used for Conan Center](https://github.com/conan-io/conan-docker-tools).

The process, described in the Linux part of `.github/workflows/ci-conan.yml`, now goes like this:
1. Generate `/tmp/osp-builder-docker/Dockerfile `, which defines the build environment as a Docker image.
2. Generate `/tmp/osp-builder-docker/entrypoint.sh`, which gets copied into the Docker image as `/entrypoint.sh` and contains the actual _libcosim_ build steps.
3. Build a Docker image named `osp-builder` based on the above files.
4. Run a Docker container based off `osp-builder`, i.e., run `/entrypoint.sh` inside the build environment. The source folder which is checked out on the host runner gets mounted read-only at `/mnt/source` inside the container.

I've included a change which was originally in #633, namely the addition of the environment variable `LIBCOSIM_RUN_TESTS_ON_CONAN_BUILD`, added so we can run tests in Conan-based CI builds too.

### DIFF
--- a/.github/workflows/ci-conan.yml
+++ b/.github/workflows/ci-conan.yml
@@ -67,6 +67,7 @@ jobs:
       CONAN_NON_INTERACTIVE: True
       CONAN_USE_ALWAYS_SHORT_PATHS: True
       CONAN_USER_HOME_SHORT: C:\c
+      LIBCOSIM_RUN_TESTS_ON_CONAN_BUILD: True
     strategy:
       fail-fast: false
       matrix:
--- a/.github/workflows/ci-conan.yml
+++ b/.github/workflows/ci-conan.yml
@@ -3,24 +3,13 @@ name: libcosim CI Conan
 # This workflow is triggered on pushes to the repository.
 on: [push, workflow_dispatch]
 
-env:
-  CONAN_LOGIN_USERNAME_OSP: ${{ secrets.osp_artifactory_usr }}
-  CONAN_PASSWORD_OSP: ${{ secrets.osp_artifactory_pwd }}
-  CONAN_REVISIONS_ENABLED: 1
-  CONAN_NON_INTERACTIVE: True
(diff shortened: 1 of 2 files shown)


kyllingstad:
@markaren, two notes for PR #633:
* You probably want to re-enable the proxy tests in your PR once it's merged with this one, i.e. remove my line 62 in `conanfile.py`.
* I changed the prefix of the new environment variable from `COSIM_` to `LIBCOSIM_` since that's the convention we've used elsewhere.

ljamt (approved):
Looking good :)

markaren (approved):

markaren:
Wupsie, seems tests are not enabled for windows (ENV not set)

kyllingstad:
Good catch! And enabling it revealed another problem: When environment variables are set to `True` in a workflow YAML file, it gets interpreted as a *boolean* rather than a string and converted to lowercase `true` during the parsing process. Conan only understands the literal values `True` and `1` for its environment variables, and `true` gets interpreted as false. :-|

This is probably an issue with the CMake CI setup too, so I'll investigate and open a separate PR for that if necessary.

markaren (approved):

PR #647, opened 2021-08-19
Bug/646 observer memory leaks

restenb:
This PR aims to fix the memory leaks in `time_series_observer.cpp` and `file_observer.cpp`. 

For `file_observer`, the `std::stringstream ss_` that was "collecting" data to write to file has been reduced in scope to a local variable in the functions that used it. 

For `time_series_observer`, the culprit was the zero argument constructor requesting a buffer size of 0, which bypassed the `adjustIfFull `test and led to constantly increasing buffer sizes for all variable types. The zero argument constructor will now instead provide a default buffer size of `10000` samples. A zero sample observer is no longer allowed and will throw, the minimum buffer size is 1. 

The changes have been tested with cosim-cli and memory use found to be stable at 5-6 MB over a 6000 second simulation of the techlaunch-tutorial configuration while logging everything. Running the same simulation on the current master branch leads to memory usage steadily climbing to 10 times that.

### DIFF
--- a/include/cosim/observer/file_observer.hpp
+++ b/include/cosim/observer/file_observer.hpp
@@ -45,7 +45,7 @@ public:
      * \param logDir the directory where log files will be created.
      * \param configPath the path to an xml file containing the logging configuration.
      */
-    file_observer(const cosim::filesystem::path& logDir, const cosim::filesystem::path& configPath);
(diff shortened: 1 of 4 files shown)


kyllingstad (changes requested):
Everything looks good, except that the default buffer size of 10000 needs to be mentioned in the documentation.

kyllingstad (approved):
LGTM

PR #648, opened 2021-08-20
Changelog for 0.8.0

ljamt:
Updating changelog with accomplishments for 0.8.0.

Before release:
#647 must be merged
proxy-fmu must be released


kyllingstad (approved):
The only minor nit I have to pick is that two bullets start with a capital letter but one doesn't. But I see we haven't been consistent in that regard earlier either, so it's probably not worth fretting over.

Issue #654, opened 2021-09-07
Simulate until allows single steps to be taken after end time is passed and "stopped" flag is not reversed if simulate_until ends by timeout


msteinsto:
Current do-loop implementation of `simulate_until()` allows single steps to be taken after end time is exceeded if the optional end time is provided. It would probably be more intuitive that the simulation will only step if the end time is not exceeded. The step function can be used if a single simulation step is the intended behavior. 

Furthermore, the `stopped_` flag is not reset if the `simulate_until()` function is ended by timeout. `is_running()` is determined by the inverse of the `stopped_` flag. The simulation then ends up in a state where it is effectively paused as it is no longer stepping while the indicated state accessed by `is_running()` will return true. `is_running()` is documented as "Is the simulation loop currently running" which it will not be if the `simulate_until()` while loop is ended by `time_out()` returning true. `stop_simulation()` is documented as "Stops the co-simulation temporarily" which should probably be called after the `simulate_until()` function has successfully completed all steps. `stop_simulation()` sets the `stopped_` flag to true.

PR #656, opened 2021-09-09
Stopped flag is now set true after simulate_until loop is finished

msteinsto:
Possible fix to one of the issues described in #654 with the `simulate_until()` function not setting the `stopped_` flag to `true` after the optional stop time has been reached, as if it was paused manually at the stop time

### DIFF
--- a/src/cosim/execution.cpp
+++ b/src/cosim/execution.cpp
@@ -151,7 +151,7 @@ public:
             do {
                 stepSize = step();
                 timer_.sleep(currentTime_);
-            }  while (!stopped_ && !timed_out(endTime, currentTime_, stepSize));
+            } while (!stopped_ && !timed_out(endTime, currentTime_, stepSize));
             bool isStopped = stopped_;
             stopped_ = true;
             return !isStopped;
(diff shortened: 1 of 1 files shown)


src/cosim/execution.cpp
```diff
@@ -151,8 +151,10 @@ class execution::impl
             do {
                 stepSize = step();
                 timer_.sleep(currentTime_);
-            } while (!stopped_ && !timed_out(endTime, currentTime_, stepSize));
-            return !stopped_;
+            }  while (!stopped_ && !timed_out(endTime, currentTime_, stepSize));
```
kyllingstad, line 154: The only change in this line seems to be an additional space which isn't supposed to be there.

kyllingstad (changes requested):
The fix looks good, but please remove the unnecessary space.

src/cosim/execution.cpp
msteinsto: Space is removed in latest commit

kyllingstad (approved):
Looks good!

PR #662, opened 2021-09-22
Use conan.cmake

### LINKED ISSUE #599: Add script for running conan install
The process of running conan correctly is not very easy. I know I have difficulties each time I clone libcosim. I suggest we add an example script to `/tools` that might look like this:

```bash
#!/usr/bin/env bash

conan remote add osp https://osp.jfrog.io/artifactory/api/conan/conan-local  --f

unameOut="$(uname -s)"
case "${unameOut}" in

    MINGW*)     conan install .. -s build_type=Debug --install-folder=../cmake-build-debug -o proxyfmu=True --build=missing
                conan install .. -s build_type=Release --install-folder=../cmake-build-release -o proxyfmu=True --build=missing;;
    Linux*)     conan install .. -s build_type=Debug -s compiler.libcxx=libstdc++11 --install-folder=../cmake-build-debug -o proxyfmu=True --build=missing
                conan install .. -s build_type=Release -s compiler.libcxx=libstdc++11 --install-folder=../cmake-build-release -o proxyfmu=True --build=missing;;
esac
```

Cavecats: 
* Builds both debug and release
* builds with proxyfmu
* Installs into CLion default directory

These options may not suit everyone, but at least it's useful as an example.

markaren:
Controversial (?) PR incoming.
Make use of conan CMake wrapper so that users only have to specify `-DLIBCOSIM_USING_CONAN=TRUE` and CMake handles everything conan related. 

Makes stuff much easier.

If OK, README should be updated as well. 

Closes #599 as this would make it unnecessary.


kyllingstad:
Controversial indeed. :) How does it make things easier?

Today, we have a situation where Conan drives the CMake build process. That is, you can run `conan create`, `conan build`, etc., and they will trigger various parts of the CMake build process.

With this proposal, it seems like we add another layer of complexity on top of this, where CMake is driving Conan, which is again driving CMake...

According to the comments in `cmake/conan.cmake` it looks like this was designed for a different use case (emphasis mine):
> It is intended to facilitate developers building projects that have conan dependencies [...]. It is not necessary to create conan packages, **in fact it shouldn't be use for that**.

This indicates it is rather designed to simplify building of "non-package" projects (like cosim-cli).

markaren:
> This indicates it is rather designed to simplify building of "non-package" projects (like cosim-cli).

That is why we do:
```
if(CONAN_EXPORTED) # in conan local cache
        # standard conan installation, deps will be defined in conanfile.py
        # and not necessary to call conan again, conan is already running
        include(${CMAKE_BINARY_DIR}/conanbuildinfo.cmake)
        conan_basic_setup()
    else()
```

markaren:
> How does it make things easier?

Everything works as before, but the user does not need to manually call conan before building the CMake project. This is a pain, that #599 tried to improve, but that PR is far from perfect.

markaren:
It should be mentioned that this PR makes it trivial to use WSL as long as #678 is fixed.

markaren:
WSL is able to compile, but can't find runtime libraries (boost)

markaren:
I still really want this PR to be merged. It makes life so much easier not having to deal manually with CMake and it makes WSL integration a breeze. I just made the feature optional, so in order to reap the benefits one should enable the CMake option `LIBCOSIM_USING_CONAN_AUTO_CONFIG`. Otherwise, everyhing will be as before.

CMakeLists.txt
```diff
@@ -98,11 +99,32 @@ set(LIBCOSIM_EXPORT_TARGET "${PROJECT_NAME}-targets")
 # ==============================================================================
 
 if(LIBCOSIM_USING_CONAN)
-    if(EXISTS ${CMAKE_BINARY_DIR}/conanbuildinfo.cmake)
-        include("${CMAKE_BINARY_DIR}/conanbuildinfo.cmake")
-        conan_basic_setup(NO_OUTPUT_DIRS)
+
+    if(NOT LIBCOSIM_USING_CONAN_AUTO_CONFIG OR CONAN_EXPORTED) # in conan local cache
```
kyllingstad, line 103: Please add some documentation that explains what happens when autoconfig is enabled. Either above this line (in which case it should describe the whole `if`/`else` block) or add a separate explanation for the `else` part below. Right now, only the default choice is explained, and that's the *least* confusing part, IMO.

kyllingstad (changes requested):
For the record: I still don't understand what's so painful about `conan install`, and I think this is complicating the build system for a marginal use case.

That said, since @markaren seems very eager to have this, and it's just a minor and localised change to the build system, I won't stand in the way. I just have a small request for an explanatory comment.

kyllingstad:
Out of curiosity, what is it about WSL that currently prevents compilation? Isn't WSL supposed to give you a full-fledged Linux instance now?

markaren:
>  I still don't understand what's so painful about conan install

I must be running this command more often than others and on multiple systems. Each time I have to spend time figuring out the command, and run for multiple builds etc. With this PR you do a load CMake. Done.

> Out of curiosity, what is it about WSL that currently prevents compilation

The difficulty comes when using WSL as a toolchain through CMake. Making conan work there without this is beyond me.

kyllingstad:
Seems like you're dealing with way more complicated setups than me. I just do `conan install path/to/source` and then it's ready for the standard CMake commands after that. I'didn't even know cross-compilation with WSL-as-a-toolchain was a thing. Sounds complicated indeed. ;)

kyllingstad (approved):

Issue #404, opened 2019-10-03
Introduce DefaultExperiment tag in OspSystemStructure.xml


eidekrist:
Let's continue the discussion from #333, where @markaren proposed to introduce a `DefaultExperiment` tag in OspSystemStructure.xml:
> My comment about putting the (optional) runtime specific entries into a `DefaultExperiment` tag were never addressed. I'd really like to see something like:
> 
> ```
> <DefaultExperiment startTime="0.0">
>       <Algorithm name="fixedStep" stepSize="1e-4"></Algorithm>
> </DefaultExperiment>
> ```
> 
> Which can be completely ommited btw, rather than:
> 
> ```
> <StartTime>0.0</StartTime>
> <BaseStepSize>1e-4</BaseStepSize>
> <Algorithm>fixedStep</Algorithm>
> ```
> 
> However, the parameters to the Algorithm would be algorithm specific, dunno how that should best be handled. Probably some annotations stuff. Then basically you can have the exact same layout in SSP.

PR #674, opened 2021-11-18
Replace fibers with threadpool

markaren:
This PR replaces #671. It uses a threadpool rather than `std::for_each` so that compilers older than gcc9 will work.

Note that slave state is unimplemented in this PR. It will eventually be included when I have figured out how to best make it back in after removing `async_slave`. 
Additionally the `concurrency` test has been commented out. The test is about file locking.


src/cosim/utility/thread_pool.hpp
```diff
@@ -0,0 +1,88 @@
...
+                auto task = work_queue_.front();
+                work_queue_.pop();
+                task();
+
+                lck.unlock();
+                cv_.notify_one();
+
+            } else {
+                std::this_thread::yield();
+            }
```
davidhjp01, line 42: Thanks for working on this PR. Would it be possible to have an option to run fmus without spawning a separate worker thread? (i.e. running everything in a main thread sequentially). For some use-cases, the communication overhead between the main and the worker threads seem quite significant.

kyllingstad: This is possible today, and was in fact the motivation for using fibers in the first place. (Not that fibers are required for sequential execution of FMU functions, but they provide a nice framework for combining sequential execution with asynchronous I/O so one can run both local and remote slaves without the overhead of threads.)

kyllingstad: The problem is just that the remote-slave implementation we're using today (proxyfmu) is not built around async I/O, so it requires a worker thread per slave anyway. That is why fibers simply are an extra overhead that comes _in addition to_ thread and I/O today, which is what I suppose motivated @markaren's PR.

As I see it, there are two ways to improve performance in this area: Drop fibers and go all-in for proxyfmu and one-thread-per-slave (as this PR proposes), or keep fibers and implement async communication with slaves (as was the original plan).

davidhjp01: Thanks for the explanation @kyllingstad.

I was able to achieve required simulation performance by removing `std::thread` for slaves as well as recursive fiber creation in the libcosim master branch. I was not able to achieve this via `pseudo_async_slave` because it still creates fiber recursively in the fmu interface methods (calling get/setters for variables and do_step).

But now the problem is I cannot not simulate some fmus because they run directly in fibers with limited resources (created by `slave_simulator::impl::do_step`). And I cannot replace/extend `slave_simulator`  as it is always added in `execution::add_slave`.

markaren: > Would it be possible to have an option to run fmus without spawning a separate worker thread?
Yes, and this is one of the motivations for this PR. IMO the master algorithm itself should decide how it handles execution. In the case of this implementation it could take `numThreads` as an argument where < 1 means no pool.

proxy-fmu was not a motivation for this PR. My motivation was to simplify the code base and make it run faster. I've found that the fiber solution is slower than a threaded solution with or without proxy-fmu.

markaren:
> It uses a threadpool rather than std::for_each so that compilers older than gcc9 will work

This is an important thing to address if this is something we want to move forward with. Do we have to support old compilers that are not C++17 feature complete?

markaren:
If the target branch gets the ok, we should also make it possible to specify the count in the XML config.

markaren:
Adding the functionality to set the number of worker threads for SSP was trivial, but not so much for the OSP alternative as it does not provide algorithm spesific configuration options. This is related to #404

ljamt:
> Adding the functionality to set the number of worker threads for SSP was trivial, but not so much for the OSP alternative as it does not provide algorithm spesific configuration options. This is related to #404

I see your point, but I don't think this can be solved separately. Should not be a blocker for this PR.

ljamt:
What's the reason for this PR to remain in draft mode? Any remaining issues that must be resolved?

markaren:
> What's the reason for this PR to remain in draft mode? Any remaining issues that must be resolved?

Not other than significant vetting.

restenb (changes requested):
Great job getting a PR out with fibers removed! 

I left some review comments, particularly around the new thread pool functionality. The implementation looks  correct, but suggestions for improvements.

src/cosim/utility/thread_pool.hpp
```diff
@@ -0,0 +1,106 @@
...
+    void worker_thread()
+    {
+        while (!done_) {
+
+            std::unique_lock<std::mutex> lck(m_);
+            if (!work_queue_.empty()) {
+
+                auto task = work_queue_.front();
+                work_queue_.pop();
+                task();
```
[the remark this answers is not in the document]
davidhjp01: This might causing the simulation (fmu steps) not parallelized since other worker threads are blocked until the mutex gets released.

src/cosim/algorithm/fixed_step_algorithm.cpp
```diff
@@ -60,8 +49,9 @@ int calculate_decimation_factor(
 class fixed_step_algorithm::impl
 {
 public:
-    explicit impl(duration baseStepSize)
+    explicit impl(duration baseStepSize, std::optional<unsigned int> workerThreadCount)
         : baseStepSize_(baseStepSize)
+        , pool_(std::min(workerThreadCount.value_or(std::thread::hardware_concurrency()), std::thread::hardware_concurrency()))
```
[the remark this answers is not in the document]
markaren: That would by default decrease the performance, which rarely would be what you want? Users that want to use less threads could always do that in configuration.

src/cosim/utility/thread_pool.hpp
```diff
@@ -0,0 +1,106 @@
...
+    }
+
+    template<typename FunctionType>
+    void submit(FunctionType f)
+    {
+        if (threads_.empty()) {
+            f();
+        } else {
+            std::unique_lock<std::mutex> lck(m_);
+            work_queue_.push(std::function<void()>(f));
```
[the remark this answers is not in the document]
markaren: FunctionType could be a named function. I'm unsure if that makes a difference.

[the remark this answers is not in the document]
markaren: Ahh, I guess the reason for writing it like that is that FunctionType does not necessary have the signature `std::function<void()>`.

src/cosim/utility/thread_pool.hpp
```diff
@@ -0,0 +1,106 @@
...
+            std::unique_lock<std::mutex> lck(m_);
+            while (!work_queue_.empty()) cv_.wait(lck);
+        }
+    }
+
+    template<typename FunctionType>
+    void submit(FunctionType f)
+    {
+        if (threads_.empty()) {
+            f();
```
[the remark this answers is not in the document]
markaren: This executes on the main thread. So, yes there is a difference.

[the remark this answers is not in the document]
markaren: The alternative would be to put this logic in `fixed_step_algorithm`

src/cosim/utility/thread_pool.hpp
```diff
@@ -0,0 +1,106 @@
...
+
+            std::unique_lock<std::mutex> lck(m_);
+            if (!work_queue_.empty()) {
+
+                auto task = work_queue_.front();
+                work_queue_.pop();
+                task();
+
+                lck.unlock();
+                cv_.notify_one();
```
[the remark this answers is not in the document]
markaren: No?

src/cosim/utility/thread_pool.hpp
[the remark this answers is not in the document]
davidhjp01: I did not create a separate PR but I see the simulation runs faster with the following modification:
```patch
Index: src/cosim/utility/thread_pool.hpp
<+>UTF-8
===================================================================
diff --git a/src/cosim/utility/thread_pool.hpp b/src/cosim/utility/thread_pool.hpp
--- a/src/cosim/utility/thread_pool.hpp	(revision HEAD)
+++ b/src/cosim/utility/thread_pool.hpp	(revision Staged)
@@ -30,23 +30,27 @@
     std::vector<std::thread> threads_;
     std::mutex m_;
     std::condition_variable cv_;
+    std::condition_variable wk_cv_;
+    uint32_t num_tasks;
 
     void worker_thread()
     {
+        std::chrono::seconds wait_duration(1);
         while (!done_) {
-
             std::unique_lock<std::mutex> lck(m_);
+
             if (!work_queue_.empty()) {
-
                 auto task = work_queue_.front();
                 work_queue_.pop();
-                task();
-
                 lck.unlock();
+                task();
+
+                lck.lock();
+                num_tasks--; // This needs to be decremented after the task is actually completed
                 cv_.notify_one();
-
             } else {
                 std::this_thread::yield();
+                wk_cv_.wait_for(lck, wait_duration);
             }
         }
     }
@@ -54,6 +58,7 @@
 public:
     explicit thread_pool(unsigned int thread_count = std::thread::hardware_concurrency())
         : done_(false)
+        , num_tasks(0)
     {
         try {
             for (unsigned i = 0; i < thread_count; ++i) {
@@ -73,7 +78,9 @@
     {
         if (!threads_.empty()) {
             std::unique_lock<std::mutex> lck(m_);
-            while (!work_queue_.empty()) cv_.wait(lck);
+            while (num_tasks > 0) {
+                cv_.wait(lck);
+            }
         }
     }
 
@@ -84,14 +91,16 @@
             f();
         } else {
             std::unique_lock<std::mutex> lck(m_);
+            num_tasks++;
             work_queue_.push(std::function<void()>(f));
+            wk_cv_.notify_one();
         }
     }
 
     ~thread_pool()
     {
         done_ = true;
-
+        wk_cv_.notify_all();
         for (auto& thread : threads_) {
             if (thread.joinable()) {
                 thread.join();

```

src/cosim/utility/thread_pool.hpp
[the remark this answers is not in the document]
restenb: The idea being to wake up one of the worker threads immediately after adding to the work queue, and in the same function, rather than waiting for one of the running threads to process the queue. 

This strikes me as a cleaner implementation that also removes the need to loop all the worker threads constantly. They can simply block until something calls `submit()`, then be notified from here.

src/cosim/utility/thread_pool.hpp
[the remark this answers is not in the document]
restenb: Adding my suggestions in code as well (may include typos since I just wrote this into the text editor);

Changes to `worker_thread()`:
```
void worker_thread()
{
    while (true)
    {
        {
            std::unique_lock<std::mutex> lck(m_);
            cv_.wait(lck, [=]{return done || !work_queue_.empty()});
    
            if (done && work_queue_.empty()) break; 

            auto task = std::move(work_queue_.front());
            work_queue_.pop();
        }
        task();
    }
}
```

Changes to `submit()`:

```
template<typename FunctionType>
void submit(FunctionType f)
{
    if (threads_.empty()) {
        f();
    } else {
       {
           std::unique_lock<std::mutex> lck(m_);
           work_queue_.emplace(std::move(std::function<void()>(f)));
       }
      cv_.notify_one();
   }
}
```

src/cosim/utility/thread_pool.hpp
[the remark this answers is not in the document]
davidhjp01: This is when `workerThreadCount=0` so we want to invoke the function directly instead of pushing it to the queue (i.e. there is no worker thread).

src/cosim/algorithm/fixed_step_algorithm.cpp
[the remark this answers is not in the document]
restenb: I'd assume you want one thread free for the whatever application you're running that embeds libcosim. 

Though the easier solution I think is to remove the reliance on defaults, and just check user input against `std::thread::hardware_concurrency()` to make sure the requested `workerThreadCount `number can actually be spawned.

src/cosim/utility/thread_pool.hpp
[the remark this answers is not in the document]
restenb: Yep, but at that point we don't really have a thread pool. Why make it responsible for running work on the main thread also? 

I think this should be moved out of the thread pool implementation.

src/cosim/utility/thread_pool.hpp
```diff
@@ -29,30 +29,28 @@ class thread_pool
...
                 lck.lock();
                 pending_tasks_--;
                 lck.unlock();
-
-                cv_.notify_one();
-
+                cv_finished_.notify_one();
             } else {
-                std::this_thread::yield();
+                cv_worker_.wait_for(lck, wait_duration);
```
restenb, line 53: I don't really agree with this. Can't we just block the thread until work is available through submit()? This looks like it'll wake the thread up once every second, but I don't immediately see a reason to do so.

restenb: That sounds like a synchronization issue? 

I think we need to use the mutex in the destructor also; 

```
std::unique_lock<std::mutex> lck(m_);
done_ = true;
cv_worker_.notify_all();
lck.unlock();
```

davidhjp01: From what I understand, `notify_all` can be missed by the worker thread if it is issued while the thread is not in the waiting state (i.e. not blocked by `wait()`). And since `notify_all` can be issued at any arbitrary time (e.g. when the program terminates), it is difficult to control. So it would be safer and cleaner for the thread just to 'timeout' periodically with a reasonable interval to check if `done_=true`.

And yes I think we can guard `done_` in the destructor.

restenb (changes requested):

src/cosim/utility/thread_pool.hpp
restenb: I committed some changes now, mainly to facilitate the following suggestions: 

* If no work is available, immediately block the worker thread via `cv_worker_`. Revive when we're either `done_`, or `work_queue_ `is not empty (don through `notify_one()` in `submit()`). 
* `wait_for_tasks_to_finish` should wait `pending_tasks_ `and the `work_queue_ ` to empty, not the threads (!)
* Increment pending_tasks_ only when the worker thread "begins" (may not make much of a difference, but seems cleaner) 
* Lock the destructor also via the mutex in order to modify `done_ `and call `notify_all()`. Since all modifications of `done_ `are now in a mutex, it can also be changed to a regular `bool`. In addition I removed the `joinable()` query - if all threads are either running or notified, do we need it? 

I left the previous code in comments for easier comparison. Please have a look.

davidhjp01: I tested the updated version with several test runs, and seems the program terminates correctly (without blocking).
Some comments:
1. Wondering if the worker thread can miss `notify_all` when it is issued while the thread is just before acquiring the lock? (`std::unique_lock<std::mutex> lck(m_)`).
2. The lock does not go out of scope at the end of `if` block unlike the comment mentioned.

src/cosim/utility/thread_pool.hpp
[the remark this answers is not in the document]
markaren: Ideally, yes. But it simplifies the logic and amount of code in the algorithm alot. The "pool" is not public API, so we could get away with it even though it breaks the prinicple of cohesion.

src/cosim/utility/thread_pool.hpp
restenb: According to the `condition_variable `documentation, it's not necessary to keep the lock when calling `notify_one `or `notify_all`, so I've changed that back to manually calling `unlock()` in both cases.

restenb: Further, the documentation details how `condition_variable `should be used in tandem with a `mutex`, both when modifying a shared variable, and waiting on the condition:

https://en.cppreference.com/w/cpp/thread/condition_variable

I'd say the code is correct now based on this, I can't really spot any deviations from these requirements.

davidhjp01: Seems `wait()` with the pred will run the pred before the thread is put into wait https://en.cppreference.com/w/cpp/thread/condition_variable/wait. Since done_ is guarded, I think this implementation is fine :)

include/cosim/slave.hpp
```diff
@@ -221,6 +221,60 @@ class slave
...
+    };
+
+    variable_values get_variables(
+        gsl::span<const value_reference> real_variables,
+        gsl::span<const value_reference> integer_variables,
+        gsl::span<const value_reference> boolean_variables,
+        gsl::span<const value_reference> string_variables) const
+    {
+
+        variable_values values;
```
davidhjp01, line 248: I think we should let the client to decide to whether to re-use this variable for improving performance. For this I made #687.

src/cosim/utility/thread_pool.hpp
```diff
@@ -0,0 +1,121 @@
...
+    std::vector<std::thread> threads_;
+    std::mutex m_;
+    std::condition_variable cv_finished_;
+    std::condition_variable cv_worker_;
+    unsigned int pending_tasks_;
+
+    void worker_thread()
+    {
+        while (true) {
+            std::unique_lock<std::mutex> lck(m_);
```
davidhjp01, line 39: Will it be possible to move this out of while loop so we don't create/destroy lck for each loop iteration?

davidhjp01 (commented):
Good work! Just added some comments

src/cosim/utility/thread_pool.hpp
restenb: No, I don't think so. How do you then reacquire the lock after a task has been completed, if it's outside of this loop? You'd have to let the thread terminate and recreate it, which we don't want?

ljamt:
`utility_concurrency_unittest` is still commented out. Should be included and fixed before merging.

ljamt:
> `utility_concurrency_unittest` is still commented out. Should be included and fixed before merging.

Test included and passing now

restenb (approved):

davidhjp01 (approved):

ljamt:
As fibers are removed, I think `cosim::utility::shared_mutex` can be replaced with `std::shared_mutex`. Don't want to add more to this PR, and can push that as a separate PR.

ljamt:
@markaren, if you are ok with the latest changes to `thread_pool.hpp` are we then ready to merge?
@kyllingstad, @eidekrist, share your opinions if you disagree :)

markaren:
How are the observed differences in usage/speed/accuracy on your side? All good?
We can probably set the the default number of threads to `std::thread::hardware_concurrency()-1` in `fixed_step_algorithm` as suggested?

restenb:
> How are the observed differences in usage/speed/accuracy on your side? All good? We can probably set the the default number of threads to `std::thread::hardware_concurrency()-1` in `fixed_step_algorithm` as suggested?

Yes. I've added a `unsigned int max_threads_ = std::thread::hardware_concurrency() - 1` variable to `fixed_step_algorithm`. Though with what is now a blocking only strategy, it may no longer be necessary. I'd like to extend this in the future to include a spinlock as well though. Blocking & resuming threads has a non-negligible overhead cost when done at a higher rate, such as with a very small timestep simulation. 

We seem to be seeing ~15-20% improvements in simulation speed with this PR over the fiber implementation.

include/cosim/algorithm/fixed_step_algorithm.hpp
```diff
@@ -30,8 +30,11 @@ class fixed_step_algorithm : public algorithm
      *
      *  \param baseStepSize
      *      The base communication interval length.
+     *
+     *  \param workerThreadCount
+     *      A number of worker threads to spawn for running FMUs
      */
-    explicit fixed_step_algorithm(duration baseStepSize);
+    explicit fixed_step_algorithm(duration baseStepSize, std::optional<unsigned int> workerThreadCount = std::nullopt);
```
kyllingstad, line 37: `optional` feels heavy-handed here. A more "C++-ish" solution would be to use a native integer type and plain zero instead of `nullopt`.

src/cosim/utility/concurrency.hpp
```diff
@@ -28,75 +28,6 @@ namespace cosim
...
-
-private:
-    std::optional<value_type> value_;
-    mutable boost::fibers::mutex mutex_;
-    boost::fibers::condition_variable condition_;
-};
-
-
 /**
  *  A shared mutex à la `std::shared_mutex`, but with support for fibers.
```
kyllingstad, line 32: I don't think we need this class at all anymore, we can just use `std::shared_mutex`.

kyllingstad (approved):
Great work! I don't have much to add except some stylistic nitpicks here and there. I didn't look at the thread pool implementation in any detail, since it seems to have been thoroughly reviewed by others. Everything else looks good to merge as far as I'm concerned.

src/cosim/utility/concurrency.hpp
kyllingstad: Nevermind, I see that @ljamt already suggested we do this in a separate PR.

restenb: Yes, we noticed this as well. In addition `cosim::utility::shared_mutex` was only used in `utility_concurrency_unittest.cpp`, with the intention to test that file locking functioned correctly with the custom mutex, which was in turn only necessary because of fibers. 

With the removal of fibers, that test can be removed as well. We wanted to push this as a separate PR to avoid more noise in this one, but if we all agree they can be removed, might as well do it here.

include/cosim/orchestration.hpp
```diff
@@ -191,7 +191,7 @@ class fmu_file_uri_sub_resolver : public model_uri_sub_resolver
  *  URI resolver).
  */
 std::shared_ptr<model_uri_resolver> default_model_uri_resolver(
-    std::shared_ptr<file_cache> cache = nullptr);
+    const std::shared_ptr<file_cache>& cache = nullptr);
```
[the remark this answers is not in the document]
markaren: Probably just a itchy trigger finger from CLions clang suggestions

include/cosim/algorithm/fixed_step_algorithm.hpp
markaren: `std::nullopt` means `std::hardware_currency-1`. 0 means 0. `std::optional` provides more information here.

kyllingstad: Ah, so the number refers to the number of _additional_ threads beyond the main thread? Then I completely agree.

kyllingstad:
One more thing: You may want to consider running clang-format on everything before merging. I see there are some includes that are out of alphabetic order after the `async_slave` --> `slave` change, and possibly other things. If it's not fixed now, it's going to show up in someone else's PR later.

PR #680, opened 2022-02-15
Passing thread_count from the ctor.

davidhjp01:
Just added a small modification to pass `thread_count` via ctor of fixed_step_algorithm. When `threads_.empty()`, thread_pool executes the function directly without queuing.


markaren:
`thread_count` should be written in pascalCase to be in-line with the existing parameter. Other than that, LGTM.

markaren:
One more thing. You'd expect that a `threadCount` of 1 means "run on the main thread". That's not what happens here.

davidhjp01:
Thanks updated accordingly.

markaren (approved):

PR #683, opened 2022-02-16
Retire VS 2016, gcc7 and gcc8

markaren:
This PR removes old compilers from the CI. It does not add new ones.
VS2016 must be removed soon anyway (deprecated in the runners). 
gcc7 and gcc8 does not support `<execution>` in C++17, which we aim to use in the future.

### DIFF
--- a/.github/workflows/ci-cmake.yml
+++ b/.github/workflows/ci-cmake.yml
@@ -21,7 +21,7 @@ jobs:
       matrix:
         os: [ubuntu-18.04]
         build_type: [Debug, Release]
-        compiler_version: [7]
+        compiler_version: [9]
         shared_libs: [ON, OFF]
 
     steps:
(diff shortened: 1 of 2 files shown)


THEvang (approved):

markaren:
I will need further confirmation that `gcc8` can be removed. We only discussed `gcc7` in the meeting.

eidekrist:
I'm OK with this, but can't comment on the need (or lack thereof) for building with GCC 8. 

If I remember correctly the rationale for supporting GCC 7 was that it shipped with Ubuntu 18.04?

eidekrist (approved):

markaren:
Ubuntu 18.04 ships with gcc 7.5. Ubuntu 20.04 ships with gcc 9.3.

davidhjp01 (approved):
LGTM!

ljamt (changes requested):
As `<execution>` will not be an issue for the threadpool solution, I suggest holding off removing gcc7 and gcc8 for a while. Lets remove VS2016 though.

PR #687, opened 2022-03-02
Some performance tweak

davidhjp01:
Following updates are for further performance improvement:
~~1.Additional placement of `LIBCOSIM_NO_FMI_LOGGING` for disabling logging in `step_finished_placeholder`~~
1. Removed `step_finished_placeholder` and assigned `nullptr` to the callback pointer.
2. Now `cosim::slave::get_variables` does not create `cosim::slave::variable_values` for each time it's invocation, instead copies data to the argument that is passed to the function.
   - This changes the use of `get_variables`, see the updated test cases.
4. In `set_variable_cache::modfiy_and_get`, `std::unordered_map<value_reference, size_t> back_refs_` is used to only update variable values that are subject to modification (via modifier). This avoids unnecessary iteration of all variables in each simulation step.


src/cosim/slave_simulator.cpp
```diff
@@ -223,18 +225,17 @@ class set_variable_cache
...
                 }
-            }
-            assert(references_.size() == values_.size());
-            for (std::size_t i = 0; i < references_.size(); ++i) {
-                const auto iterator = modifiers_.find(references_[i]);
-                if (iterator != modifiers_.end()) {
-                    values_[i] = iterator->second(values_[i], deltaT);
+                const auto it = back_refs_.find(ref);
+                if (it != back_refs_.end()) {
+                    values_[it->second] = entry.second(values_[it->second], deltaT);
```
eidekrist, line 235: I'm wondering about the need for `back_refs_`. Isn't the array index (`it->second` here) stored in `exposedVariables_.at(ref).arrayIndex`?

eidekrist (commented):
It's been a long time since I've looked at this code :) The changes in the `fmu.cpp`classes LGTM, I've left a comment in `slave_simulator.cpp`.

src/cosim/fmi/v1/fmu.cpp
```diff
@@ -166,9 +166,11 @@ namespace
 {
 void step_finished_placeholder(fmi1_component_t, fmi1_status_t)
 {
+#ifndef LIBCOSIM_NO_FMI_LOGGING
```
markaren, line 169: `LIBCOSIM_NO_FMI_LOGGING` was introduced to supress logging comming from the FMUs itself, but this supresses libcosim generated output. I'm not sure that is a good idea.

davidhjp01: I think this function is assigned to the fmi callback instance (`fmi(1|2)_callback_functions_t`), which is called by the fmu?

src/cosim/slave_simulator.cpp
davidhjp01: I think you are right, adjusted accordingly.

src/cosim/slave_simulator.cpp
```diff
@@ -227,14 +227,10 @@ class set_variable_cache
                     references_.emplace_back(ref);
                     values_.emplace_back(exposedVariables_.at(ref).lastValue);
                 }
+                const auto index = exposedVariables_.at(ref).arrayIndex;
```
eidekrist, line 230: Since we are talking performance;
`exposedVariables_.at(ref)` is called *at least* two times in this loop. It might make sense to define it as a local variable before line 224.

davidhjp01: Thanks, updated accordingly.

src/cosim/fmi/v1/fmu.cpp
markaren: It is still not connected to the logging part. So IMO, the macro is taking liberties here. However, I think it is safe to remove the current logic altogether. We are not using, nor planning to use this feature.

ljamt: Shouldn't we define the callback as it can be called by FMUs? We can of course remove the logging and leave a comment describing it's existence.

markaren: stepFinished callback can be NULL per the standard.

markaren: Actually it SHOULD be NULL:
`Optional call back function to signal if the computation of a communication step of a cosimulation slave is finished. A null pointer can be provided. In this case the master must
use fmiGetStatus(..) to query the status of fmi2DoStep. If a pointer to a function
is provided, it must be called by the FMU after a completed communication step.`

ljamt: Good catch! For references this is stated in section 2.1.5 (p 21) in the FMI 2.0.1.
Let's use a null pointer.

davidhjp01: Updated

ljamt (approved):

include/cosim/slave.hpp
```diff
@@ -238,25 +238,26 @@ class slave
         std::vector<std::string> string;
     };
 
-    variable_values get_variables(
+    void get_variables(
+        variable_values* values,
         gsl::span<const value_reference> real_variables,
         gsl::span<const value_reference> integer_variables,
         gsl::span<const value_reference> boolean_variables,
-        gsl::span<const value_reference> string_variables) const
```
kyllingstad, line 245: Why was `const` removed here?

kyllingstad (approved):

PR #691, opened 2022-03-22
thread_pool update

davidhjp01:
To fix the thread-worker parallelization.


markaren (approved):
Looks good as far as I can tell.

PR #697, opened 2022-04-26
Proxyfmu no boost depedency

markaren:
Boost sometimes causes issues when linking, so I made a version of proxyfmu without boost. This needs to be tested. I.e. Does it work the same as before w.r.t file paths etc.


markaren:
Closing while upstream changes are split into separate PRs

PR #700, opened 2022-05-04
Not finding a variable in the model description is not exceptional

### LINKED ISSUE #669: Not finding a variable in the model description is not exceptional
Searching for a variable, but not finding it is not exceptional nor an error. I think it would be clearer to return an optional value containing a variable description

https://github.com/open-simulation-platform/libcosim/blob/22a89c439970a2e2588ed5e90082e6e892274aeb/src/cosim/model_description.cpp#L21

markaren:
Possible solution to #669. Involves API breakage.

Closes #669

### DIFF
--- a/src/cosim/observer/file_observer.cpp
+++ b/src/cosim/observer/file_observer.cpp
@@ -448,12 +448,16 @@ file_observer::simulator_logging_config file_observer::parse_config(const std::s
                             const auto variableDescription =
                                 find_variable(simulator->model_description(), name);
 
-                            switch (variableDescription.type) {
+                            if (!variableDescription) {
+                                throw std::invalid_argument("Can't find variable descriptor with name " + name + " for model with name " + simulator->model_description().name);
+                            }
+
+                            switch (variableDescription->type) {
                                 case variable_type::real:
                                 case variable_type::integer:
                                 case variable_type::boolean:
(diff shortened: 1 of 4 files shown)


src/cosim/observer/file_observer.cpp
```diff
@@ -448,12 +448,16 @@ file_observer::simulator_logging_config file_observer::parse_config(const std::s
                             const auto variableDescription =
                                 find_variable(simulator->model_description(), name);
 
-                            switch (variableDescription.type) {
+                            if (!variableDescription) {
+                                throw std::invalid_argument("Can't find variable descriptor with name " + name + " for model with name " + simulator->model_description().name);
```
kyllingstad, line 452: This shouldn't be an `invalid_argument` exception, as that indicates a programming error. (It was the wrong type to use before too, but even more so now.)

Ideally, we should add a (somewhat re-usable) error code like `name_not_found` to `cosim::errc` and then throw a `cosim::error` with this code. However, this would require changes in both `exception.hpp` and `exception.cpp`, and we're not consistent in our use of this exception type as it is. So I'm satisfied if `invalid_argument` just gets changed to `runtime_error`.

kyllingstad (changes requested):

kyllingstad (approved):

PR #704, opened 2022-06-23
Correct or remove minimum version specs in `README.md`

kyllingstad:
Pretty much all of them were wrong or misleading. It's very hard to remember to keep these up to date, and it's also difficult to know exactly which versions will work. So I suggest that we specify just the compiler versions we're testing in CI and leave the other tool versions unspecified.  (The minimum CMake version has to be specified in `CMakeLists.txt`, though.)


markaren (approved):

kyllingstad:
Merged despite failed CI checks, as they cannot possibly be related to this change.

PR #708, opened 2022-07-07
Implement programmatic log config

### LINKED ISSUE #701: file_observer should be programmatically configurable
Currently, the behaviour of `file_observer` can only be modified by passing it the path to a text file that contains the configuration. Those settings should also be available programmatically. Client code should not have to create a temporary file just for the purpose of controlling `file_observer`.

markaren:
Enable csv logging to be programatically configured.

Closes #701


eidekrist (approved):
Looks good, pending build status

PR #709, opened 2022-08-13
Used early return to simplify the resolve_reference method from URI. 

codemascot:
When studying the codebase the current conditional structure of this `resolve_reference` function seemed a bit complex to me. Therefore, here I'm proposing new simplified conditional structure using early return.

**NB. 100% of all the 26 tests has been passed! :slightly_smiling_face:** 

Selection_009


eidekrist:
Do we need a test case which covers the changes in code?

codemascot:
You are right @restenb.  As the original one is an exact implementation from RFC 3986, therefore this shouldn't be refactored. :+1:

PR #710, opened 2022-08-17
parseURI null check and error handling implemented

codemascot:
Although, only null check was suggested in the comment, implemented error handling too. And used the same message for both error handling and null check, which could be a bit confusing sometime. Feedback on these from the maintainers would be quite helpful.

NB. All the tests passed though.
Selection_010

### DIFF
--- a/src/cosim/osp_config_parser.cpp
+++ b/src/cosim/osp_config_parser.cpp
@@ -606,10 +606,22 @@ struct extended_model_description
         const auto xerces_cleanup = final_action([]() {
             xercesc::XMLPlatformUtils::Terminate();
         });
+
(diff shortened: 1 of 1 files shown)


restenb:
The error handler seems to be in place on the master branch already, though I don't recall which PR it was or when. In other words, the line 

`parser->getDomConfig()->setParameter(xercesc::XMLUni::fgDOMErrorHandler, &errorHandler);`

is already there in current master and doesn't need to be duplicated here. 

The error handler implementation seems to do it's own logging, so I'm not sure checking for `errorHandler.failed()` at the end is necessary. 

The `null` check is fine probably, though I'm not familiar enough with Xerces to know when the parser would return `null`. So from where I stand at least I'd only keep the `null` check from this PR.

restenb (approved):

PR #716, opened 2022-09-13
Refactor logging interface and use spdlog

markaren:
This PR refactors the logging interface so that the implementation is kept hidden.
It also replaces `boost::log` with `spdlog` and introduces `fmt` for formatting.

`fmt` can be replaced by `std` if we ever target `C++20`.

The CI has some issues, but nothing that can't be fixed.

Together with #714, the dependency on boost is greatly reduced (which may or may not be a goal).

Overall, this PR provides a slimmer more accessible API, whilst providing an interface that can change backend without breaking API compat.

Edit: `Boost::log` can be used over `spdlog` if there is consensus on that.

Please discuss..


markaren:
Version using `boost::log`: https://github.com/open-simulation-platform/libcosim/tree/boost_log

kyllingstad:
As I commented in #714, I don't see it as a goal to reduce our dependency on Boost, especially when it comes at the cost of introducing new dependencies (*two* new dependencies in this case).

I do agree that it is a worthy goal to keep our public API as free of dependencies as possible, though, and that includes Boost. At the same time, I think it's beneficial for client code to be able to hook into and use the same logging system as libcosim.  That's what cosim-cli does, for example. And client code that uses some other logging framework might still want to hook into Boost.Log to redirect libcosim's logging into its own system.

So my suggestion would be to keep Boost.Log as our logging framework, but make its use through the public API optional in some way. This could be done by restricting the use of `boost` symbols to headers that client code is not forced to include, and provide a framework-agnostic way to control the logging for client code that doesn't want to deal with Boost at all.

PR #721, opened 2022-11-04
Adding support for end time

### LINKED ISSUE #720: Support for endTime in OspSystemStructure
Rationale: `endTime` is a useful field to have in OspSystemStructure in some cases, e.g. automatic "chaining" of simulations where the overall system is only configurable if the `endTime` of each individual simulation can be specified.

restenb:
This PR adds support for parsing simulation end time out of an OspSystemStructure.xml and into the simulation configuration. 

Slightly modified osp_config_parser_test to facilitate testing of this addition. 

Closes #720


include/cosim/osp_config_parser.hpp
```diff
@@ -27,6 +27,9 @@ struct osp_config
     /// The default start time for a simulation
     time_point start_time;
 
+    /// The (optional) end time for a simulation
+    time_point end_time;
```
kyllingstad, line 31: Should this be `std::optional<time_point>`, or are you using zero to indicate "not specified"?

restenb: Changed to `std::optional` here. 

I'm not really into where to use and not use `std::optional` in APIs, but I think it makes sense here since it is optional in the XML. I kept it as a `double` in the `SimulationInformation` struct.

src/cosim/osp_config_parser.cpp
```diff
@@ -926,8 +933,16 @@ osp_config load_osp_config(
         throw std::invalid_argument(oss.str());
     }
 
+    if (simInfo.startTime > simInfo.endTime) {
+        std::ostringstream oss;
```
kyllingstad, line 937: It looks to me like this error handler will be triggered if an end time is not specified in the configuration file, because then `simInfo.endTime` will still be zero. Or am I misreading this?

restenb: You are right of course :facepalm: 

Seems making `endTime ` `std::optional` everywhere is the way to go here then. I also added a separate `config_end_time_test` to test a system structure with/without `endTime` configured.

kyllingstad (approved):

eidekrist (approved):
LGTM!

Issue #717, opened 2022-09-24
build error : "BOOST_ALL_NO_LIB" redefined


tamago117:
I'm trying `How to build` section. But I face an error at `Step 3: Build and test`
```
$ conan install .. --build=missing
$ cmake .. -DLIBCOSIM_USING_CONAN=TRUE -DCMAKE_BUILD_TYPE=Debug


$ cmake --build .
Scanning dependencies of target osp-xsd-embedder
[  1%] Building CXX object tools/osp-xsd-embedder/CMakeFiles/osp-xsd-embedder.dir/osp-xsd-embedder.cpp.o
[  2%] Linking CXX executable ../../output/debug/bin/osp-xsd-embedder
[  2%] Built target osp-xsd-embedder
[  3%] Generating ../generated/osp_system_structure_schema.hpp
 Embedding: /home/tamago/workspace/libcosim/data/xsd/OspSystemStructure.xsd -> /home/tamago/workspace/libcosim/build/generated/osp_system_structure_schema.hpp
Scanning dependencies of target cosim
[  4%] Building CXX object src/CMakeFiles/cosim.dir/cosim/algorithm/fixed_step_algorithm.cpp.o
<command-line>: error: "BOOST_ALL_NO_LIB" redefined [-Werror]
<command-line>: note: this is the location of the previous definition
cc1plus: all warnings being treated as errors
make[2]: *** [src/CMakeFiles/cosim.dir/build.make:68: src/CMakeFiles/cosim.dir/cosim/algorithm/fixed_step_algorithm.cpp.o] error 1
make[1]: *** [CMakeFiles/Makefile2:242: src/CMakeFiles/cosim.dir/all] error 2
make: *** [Makefile:141: all] error 2
```

### enviroment

- ubuntu 20.04
- gcc version 9.4.0
- cmake version 3.16.3

PR #723, opened 2022-11-22
Use gcc9 and ubuntu-20.04

### LINKED ISSUE #719: The Ubuntu-18.04 environment is deprecated and will be removed on December 1st, 2022
CI needs to be updated

markaren:
As discussed in the OSP meeting, removing gcc7.

Closes #719 

Mhhh.. This triggered #717

### DIFF
--- a/.github/workflows/ci-cmake.yml
+++ b/.github/workflows/ci-cmake.yml
@@ -19,9 +19,9 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [ubuntu-18.04]
+        os: [ubuntu-20.04]
         build_type: [Debug, Release]
-        compiler_version: [7]
+        compiler_version: [9]
         shared_libs: [ON, OFF]
     timeout-minutes: 35
 
--- a/.github/workflows/ci-cmake.yml
+++ b/.github/workflows/ci-cmake.yml
@@ -26,7 +26,7 @@ jobs:
     timeout-minutes: 35
 
     steps:
-      - uses: actions/checkout@v2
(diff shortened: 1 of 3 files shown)


restenb:
I can look at this a bit now, since we would really like a new release. Once this is fixed I can organize a release of libcosim & dependent tools as well. 

I'm not really that much into C++ build problems, so can't guarantee I'll be able to fix it... @kyllingstad, do you have any time to spare for this in the coming days?

davidhjp01:
For the error
image

I see in our cmake file
```cmake
target_compile_definitions(cosim PUBLIC "BOOST_ALL_NO_LIB=0" "BOOST_CONFIG_SUPPRESS_OUTDATED_MESSAGE=1")
```

Wondering whether the intention of "BOOST_ALL_NO_LIB=0" here is to turn on or off this feature (by setting it to zero)? Because it seems boost just checks the definition of the macro not its value.

Also it seems FindBoost adds `BOOST_ALL_NO_LIB` to the build, see the below (duplicated definitions):
image

So my suggestion is to remove "=0" part in the cmake file if our intention is to _enable_ "BOOST_ALL_NO_LIB". At least this resolved the issue when I built libcosim locally.

kyllingstad:
Good catch, @davidhjp01. It makes no sense to set `BOOST_ALL_NO_LIB=0`; the macro only exists as a means to _disable_ autolinking. But I suggest we just remove the  `=0` part as you suggest, because earlier `FindBoost` versions might not set it automatically.

kyllingstad (approved):

Issue #712, opened 2022-09-01
Signal name when logging signals to csv file


tomarnepedersen:
When signals are logged to csv file, the naming convention seem to be:
[signal name] [[signal id] [signal type]
e.g. azimuth_thrust [6 real output]

When new signals are added to the model, the id may change, even though the name is the same. So after the model has been updated and compiled to fmu, the signal could have gotten a new id, which gives e.g. the following signal name in the csv file:
azimuth_thrust [8 real output]

Then, when reading the csv file, and plotting signals, the code for plotting will need to be updated since the ID has changed. I would suggest to not include the ID in the signal name when logging to file.

eidekrist:
Moved to `libcosim` and added "discussion needed" label.

Has anyone ever used the additional metadata (`[valuereference type causality]`) for anything?

markaren:
In the very specific cases where this would be an issue, one could just ignore the metadata in the plotter. This sounds like an easy to fix issue in the plotting code. 

Altough, there probably isn't that much value in keeping `valuereference` as you'd probably search using the name anyway.

kyllingstad:
I think we should drop the additional metadata from the header line and just have the variable name there. In the few cases where such information is needed, it can be gotten from elsewhere.

> In the very specific cases where this would be an issue, one could just ignore the metadata in the plotter.

This presumes that the user has programmatic control of the plotting. But I'd argue that's not the typical case; most will use point-and-click solutions (e.g. Excel).

restenb:
This has (hopefully) been covered by https://github.com/open-simulation-platform/libcosim/pull/725 in the latest release, so I'm closing this.

PR #725, opened 2022-11-29
Separate variable reference, type, causality into separate header lines 

restenb:
This PR proposes to change the `variable_name [ref, type, causality]` pattern in the current file logging implementation to one new line for each. This might be easier to use when parsing the CSV files. 

In other words the values will now be logged starting from line 5, and the new CSV files look like this:

> 1. Time,StepCount,realOut,intOut,booleanOut,stringOut
> 2. #,#,0,0,0,0 
> 3. #,#,real,integer,boolean,string
> 4. #,#,output,output,output,output
> 5. 0,0,1.234,1,0,hello log
> 6. 0.1,1,1.234,1,0,hello log
>

### DIFF
--- a/src/cosim/observer/file_observer.cpp
+++ b/src/cosim/observer/file_observer.cpp
@@ -131,6 +131,8 @@ public:
     }
 
 private:
+    int keyWidth_ = 14;
+
     template<typename T>
     void write(const std::vector<T>& values, std::stringstream& ss)
     {
@@ -233,12 +235,27 @@ private:
         }
     }
 
+    void write_variable_metadata(std::stringstream& ss, std::vector<variable_description>& variables)
+    {
+        for (const auto& v : variables) {
+            ss << "  - " << std::setw(keyWidth_) << "name:" << v.name << std::endl
+               << "    " << std::setw(keyWidth_) << "reference:" << v.reference << std::endl
+               << "    " << std::setw(keyWidth_) << "type:" << v.type << std::endl
+               << "    " << std::setw(keyWidth_) << "causality:" << v.causality << std::endl
+               << "    " << std::setw(keyWidth_) << "variability:" << v.variability << std::endl;
+
+            if (v.start.has_value()) {
+                ss << "    " << std::setw(keyWidth_) << "start value:";
+                std::visit([&](const auto& val) { ss << val << std::endl; }, v.start.value());
+            }
+        }
+    }
+
     void create_metadata_file()
     {
         std::ofstream metadata_fw;
         std::string filename;
         std::stringstream ss;
-        constexpr int keyWidth = 14;
 
         if (!timeStampedFileNames_) {
             filename = observable_->name().append("_metadata.yaml");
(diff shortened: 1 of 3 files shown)


markaren:
Not sure I agree on this one. `realOut [ref type casualty]` is easy enough to parse, even without regex. 
I also think some parsers like pandas in python asumes a header on the first line followed by data?

I don't really see the problem to be solved here. #712 raised an issue related to `ref` which could change over time. That one can be solved in libcosim by removing the `ref` identifier as it is redundant anyway.

kyllingstad:
Me neither. It's kind of a de facto standard for CSV to have at most one header line. Not all CSV parsers will support skipping, let alone parsing, multiple header lines. I think we should simply drop the metadata from the header (as I also just commented in #712).

restenb:
Good input! I originally added `[ref, type, causality]` after the variable name just to somehow squeeze all information on one line, but it always felt awkward. I assume that most ready-to-use plotting tools assume it's just the name of the signal/axis as well. 

So I have removed the extra header lines again for now. We can provide the other information in a separate metadata file perhaps?

kyllingstad:
Yeah. `cosim` provides the information in machine-readable form (YAML, specifically), but only for individual FMUs. You can see an example under `cosim inspect` here: https://open-simulation-platform.github.io/cosim

restenb:
Nice. I looked at the implementation of `cosim inspect` and decided to bring it over to the file logger as well, so now this branch will output a `logFileName_metadata.yaml` file alongside the CSV, with the contents exactly the same as what `cosim inspect` puts on stdout.

You can have a look with `file_observer_logging_test` or any of the other tests that log to file. If anybody is parsing the old format it'll be a breaking change of course, but I think this is a better way of separating the simulation results from it's metadata. 

I'll be looking to release libcosim in the coming days - hopefully this week - so we need to decide on whether this is to be included in that as well.

kyllingstad (approved):
Nice!

src/cosim/observer/file_observer.cpp
```diff
@@ -229,6 +233,59 @@ class file_observer::slave_value_writer
...
+        ss << std::left
+           << std::setw(keyWidth) << "name:" << md.name << std::endl
+           << std::setw(keyWidth) << "uuid:" << md.uuid << std::endl
+           << std::setw(keyWidth) << "description:" << md.description << std::endl
+           << std::setw(keyWidth) << "author:" << md.description << std::endl
+           << std::setw(keyWidth) << "version:" << md.version << std::endl;
+
+        ss << "variables:" << std::endl;
+
+        for (const auto& v : md.variables) {
```
eidekrist, line 270: Instead of writing metadata for all variables, shouldn't this file reflect the content in the log file? In other words, use the logConfig information to only write data for logged variables?

eidekrist (commented):
A metadata file is a very nice solution!

src/cosim/observer/file_observer.cpp
restenb: Absolutely, good catch. Updated accordingly now. A couple runs of `file_observer_logging_from_config_test` confirms that it works and metadata is only generated for configured variables.

restenb: I also realized that in order for the CSV and YAML files to be "linkable", we have to do something, otherwise there would be no way to tell which metadata file corresponded to which CSV file when e.g. running multiple logging sessions through the same file observer. The simplest solution seems to be to just return the timestamp from create_log_file() and reuse it in create_metadata_file(), so the filenames output will look like this:

> slave uno_20221201_140801_676545.csv
> slave uno_20221201_140801_676545_metadata.yaml 

Alternatives would be creating a new directory for each logging, or linking them by specifying the CSV filename itself as metadata, but I think this solution should suffice.

PR #728, opened 2022-12-08
preparing for 0.10.1 release

restenb:
As discussed, we need a quick patch release now to proxyfmu 0.3.0 that includes FindThrift.cmake. 

I assume this removes the need for `cmake_find_package `generator both here & downstream, so I updated that. 

We can test with `proxyfmu/0.3.0@osp/testing-undo_cmake_find_package` and update once proxyfmu is released.


markaren:
Additional cmake include paths can be removed as in https://github.com/open-simulation-platform/libcosim/pull/727

markaren (approved):

Issue #718, opened 2022-09-27
Conan v2 combability


markaren:
Recipes across the OSP organization are not conan v2 compatible and will stop working at some point. They should be updated sooner than later.

https://docs.conan.io/en/latest/migrating_to_2.0/recipes.html

kyllingstad:
I'll do this as a precursor to #741.

@joakimono currently has a Conan 2 recipe for FMI library under review for conan-center; see conan-io/conan-center-index#20256. When that is in, we'll work towards packaging FMI Library 3.x too (currently in alpha). We can then start working in parallel on FMI 3.0 support in libcosim.

PR #747, opened 2023-11-20
Upgrade to Conan 2

kyllingstad:
This fixes issue #718, “Conan v2 compatibility”. (Note that it also *destroys* Conan 1.x compatibility – it was simply too difficult to maintain support for both major versions.)

Some collateral changes warrant further explanation:

* The modified `<gsl/util>` includes and various changes from `int` to `std::size_t` are a consequence of using a newer version of MS GSL. I changed the dependency version so we can use an upstream CMake package configuration file which is compatible with the one generated by Conan 2's `CMakeDeps` generator. This should enable us to use the same `find_package()` commands and target names regardless of whether the build is running under Conan or not.
* I've modified the “non-Conan” CI workflow to build in a clean Docker container like the Conan CI workflow. I've based the container on Debian Bookworm, whose packages for our dependencies are new enough that all provide CMake package configuration files which match those generated by `CMakeDeps`, which was my main motivation. The MS GSL package mentioned in the previous point is one such. (Also, the GitHub runners are such a noisy and unpredictable build environment that I figured this would be more efficient than try to fix compatibility issues there.)
* I've had to remove the option to run Conan from within CMake, because upstream hasn't released a version of the cmake-conan helper script with full support for all Conan 2 features yet. (Notably, support for the `CMakeToolchain` generator is missing.)

The pull request depends on open-simulation-platform/proxy-fmu#92, which should be merged first. Then, the `osp/testing-…` channel should be replaced in the `proxyfmu` requirement specification in `conanfile.py` before merging the present PR.

Incidentally, this PR also fixes issue #702.


kyllingstad:
Seems like this PR closes several build-related issues. I linked them now.

conanfile.py
```diff
@@ -1,91 +1,120 @@
...
+    def requirements(self):
+        self.tool_requires("cmake/[>=3.19]")
+        self.requires("fmilibrary/[~2.3]")
+        self.requires("libzip/[>=1.7 <1.10]") # 1.10 deprecates some functions we use
+        self.requires("ms-gsl/[>=3 <5]", transitive_headers=True)
+        if self.options.proxyfmu:
+            self.requires("proxyfmu/0.3.2@osp/testing-feature_conan-2")
+            self.requires("boost/[~1.81]", transitive_headers=True) # Required by Thrift
+        else:
+            self.requires("boost/[>=1.71]", transitive_headers=True)
```
davidhjp01, line 48: It seems a consumer of `libcosim` needs to explicitly declare `boost` to use, for example at least `boost::timer` in its code. Otherwise linking fails.

I probably not fully understand the "traits" from the conan doc yet, but may be `transitive_libs` allows the consumers to link boost declared here? So that "downstream" (e.g. libcosim consumers) can also use the same boost version in their code, without accidentally declaring mismatching boost version.

conanfile.py
```diff
@@ -1,91 +1,120 @@
...
 
     def configure(self):
-        self.options["boost"].shared = self.options.shared
-        self.options["fmilibrary"].shared = self.options.shared
-        self.options["libzip"].shared = self.options.shared
-        self.options["yaml-cpp"].shared = self.options.shared
-        self.options["xerces-c"].shared = self.options.shared
+        if self.options.shared:
+            self.options.rm_safe("fPIC")
+        self.options["*"].shared = self.options.shared
```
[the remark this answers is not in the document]
kyllingstad: And it is correct, it was taken from an example on this page of the Conan docs: https://docs.conan.io/2.0/reference/conanfile/methods/configure.html

conanfile.py
kyllingstad: The way I see it, consumers who use Boost directly in their own code *should* list it as an explicit dependency and not rely on the indirect dependency from libcosim. However, if they only need Boost to be able to link with libcosim, then it should be handled automatically. I'm actually not sure whether `transitive_libs` is required in that case; I'll try to find out.

kyllingstad: I *think* the reason I specified `transitive_headers` is because libcosim headers include some Boost headers, so client code must have them on the include path.

davidhjp01:
@kyllingstad @restenb 
- Wondering if we should remove Debug builds because it seems the case for the most of the libraries in conan2 public repository (to avoid lengthy github action builds).
- Also for some reason conan2 tries to rebuild some of the libraries for the Release builds, i guess due to some mismatching conan2 options?

davidhjp01:
I have addressed the most of the issues related to the failing builds. I suggest to merge and release libcosim and proxyfmu so that other projects dependent on conan2 library can consume them.

davidhjp01 (approved):

conanfile.py
kyllingstad: It turns out you were right about this, @davidhjp01. Since we use Boost in some libcosim headers, and more specifically *Boost libraries that have binary components* (i.e., are not header-only), such as Boost.Log, then we need to have `transitive_libs=True` here. Otherwise, client code won't have access to those binary components for linking. I'll make a new pull request.

kyllingstad: Should be resolved by #753.

kyllingstad:
Following up on your question (too late for this PR, unfortunately):

> Wondering if we should remove Debug builds because it seems the case for the most of the libraries in conan2 public repository (to avoid lengthy github action builds).

Yes, I think that sounds like a good idea. I don't think I can remember a case where the release builds succeeded while the debug builds failed or vice versa.

PR #753, opened 2024-02-22
Conan: Use transitive_libs=True for Boost

kyllingstad:
Since we use Boost in some libcosim headers, and more specifically Boost libraries that have binary components (i.e., are not header-only), such as Boost.Log, then we need to specify the `transitive_libs=True` trait when listing the requirement. Otherwise, client code won't have access to those binary components for linking.

Thanks to @davidhjp01 for [pointing out](https://github.com/open-simulation-platform/libcosim/pull/747#discussion_r1417957294) the issue.


eidekrist (approved):
LGTM!

davidhjp01 (commented):
Looks good!

Issue #619, opened 2020-10-22
UNC config paths not supported


eidekrist:
Loading a config with for instance an OspSystemStructure.xml located on a server drive, such as

`\\my.company\user\configuration`

results in an error with the following message:

`Could not create execution from OspSystemStructure.xml file: No resolvers available to handle URI: MyModel.fmu`

kyllingstad:
I've looked into this and figured out why it happens.

The system structure file path `\\my.company\user\configuration\OspSystemStructure.xml` becomes the file URI `file://my.company/user/configuration/OspSystemStructure.xml`. This is then used as the base URI for resolving model URI references, so `MyModel.fmu` becomes `file://my.company/user/configuration/MyModel.fmu`.

The problem is that `fmu_file_uri_sub_resolver::lookup_model()` [deliberately rejects](https://github.com/open-simulation-platform/libcosim/blob/02e1d8cf736652cd1b4fa57a2a953ecb160f4d45/src/cosim/orchestration.cpp#L121) file URIs with a non-empty *authority* field (`my.company` in this case). The reason is that the interpretation of such a URI is not well defined. The file URI scheme simply says that the authority part should contain the fully qualified domain name of the host on which the file is located, but it leaves the access mechanism undefined. So the file could be on a network share, an FTP server, etc.

However, the standard interpretation on Windows is as a network share, which was the actual case presented here. So the fix I have in mind now is to remove the authority check on Windows and use this interpretation.

PR #755, opened 2024-02-26
Allow use of UNC paths on Windows

kyllingstad:
This fixes issue #619.

File URIs generally have the form `file://hostname/path/to/file`. When `hostname` is not `localhost` (nor empty) the file URI scheme leaves the file access mechanism unspecified. However, on Windows, such URIs are commonly interpreted as UNC paths, so in this case it would be `\\hostname\path\to\file`. This PR simply enables us to take advantage of this interpretation, which is built into the Windows API functions we already use.


eidekrist (approved):
I'm not in a position to test this myself, but LGTM!

kyllingstad:
The original issue (#619) is difficult to recreate in unittests too, but I did test locally on my machine that this PR solves it. (That is, I was able to run a simulation with an UNC path to `OspSystemStructure.xml`.)

PR #758, opened 2024-04-30
Sampling period for computing the rolling average real time factor

davidhjp01:
When the simulation step size is too small, e.g. 0.001, a fixed `cosim::real_time_config.steps_to_monitor` value would not be sufficient to compute the rolling average real time factor because of the following scenario:

1. Set `steps_to_monitor` to a large value (1000) because the default 5 is too small window size to accurately compute RTF.
2. Set the simulation step size to 0.001. The solution is good enough when the simulation can keep RTF 1.
3. When RTF drops, for ex. to 0.01, during the simulation for some reason, user would only see the updated RTF after 100 seconds.

To avoid this problem, `cosim::real_time_config.sampling_period_to_monitor` is introduced. It sets a sampling period in second(s) used in the rolling average real time factor calculation. When the value is greater than zero, the real time factor is computed periodically using the specified time instead of the `steps_to_monitor` value.


conanfile.py
```diff
@@ -40,7 +40,7 @@ def requirements(self):
         self.tool_requires("cmake/[>=3.19]")
         self.requires("fmilibrary/[~2.3]")
         self.requires("libzip/[>=1.7 <1.10]") # 1.10 deprecates some functions we use
-        self.requires("ms-gsl/[>=3 <5]", transitive_headers=True)
+        self.requires("ms-gsl/[>=3 <5]", transitive_headers=True, transitive_libs=True)
```
kyllingstad, line 43: Why is this necessary? As far as I know, MS GSL is a header-only library.

include/cosim/timer.hpp
```diff
@@ -36,6 +36,15 @@ struct real_time_config
...
     std::atomic<int> steps_to_monitor = 5;
+
+    /**
+     *  A sampling period in second(s) used in the rolling average real time factor calculation.
+     *  This can be useful when simulation step size is small and a fixed value for `steps_to_monitor`
+     *  would not compute an accurate rolling average real time factor.
+     *  When the value is greater than zero, the real time factor is computed periodically using the
+     *  specified time instead of the `steps_to_monitor` value.
+     */
+    std::atomic<double> sampling_period_to_monitor = -1;
```
kyllingstad, line 47: Have you considered using a `std::chrono::duration` type for this, for example `std::chrono::milliseconds` (or `std::chrono::duration<double>` if it really needs to be floating-point)? I think it would express the semantics better, though I don't know whether it will work with `std::atomic`.

src/cosim/timer.cpp
```diff
@@ -37,6 +36,12 @@ class real_time_timer::impl
         const auto newHash = std::hash<real_time_config>()(*config_);
         if (newHash != configHashValue_) {
             start(currentTime);
+            auto samping_period = config_->sampling_period_to_monitor.load();
```
kyllingstad, line 39: Typo: `samping_period` should be `sampling_period`. Also, the variable should be `const`.

include/cosim/timer.hpp
```diff
@@ -36,6 +36,15 @@ struct real_time_config
      * This value is used for monitoring purposes only.
      */
     std::atomic<int> steps_to_monitor = 5;
+
+    /**
+     *  A sampling period in second(s) used in the rolling average real time factor calculation.
```
kyllingstad, line 41: Perhaps you should explicitly state that the sampling period is specified in real (wall clock) time rather than logical (simulation) time, to avoid confusion.

src/cosim/timer.cpp
```diff
@@ -72,21 +77,36 @@ class real_time_timer::impl
...
     std::shared_ptr<real_time_metrics> metrics_;
+    std::optional<duration> sampling_period_to_monitor_ = std::nullopt;
+
+    void update_rolling_average_real_time_factor(
+        Time::time_point& currentTime,
+        time_point& currentSimulationTime,
+        const duration& elapsedRealTime)
+    {
+        const auto elapsedSimTime = currentSimulationTime - rtSimulationStartTime_;
+        metrics_->rolling_average_real_time_factor = elapsedSimTime.count() / (1.0 * elapsedRealTime.count());
```
kyllingstad, line 88: `elapsedSimTime` and `elapsedRealTime` are two different `std::chrono::duration` types associated with different clocks. The former is associated with `Time`, which is an alias for `std::chrono::steady_clock` and represents real time, while the latter is associated with `cosim::detail::clock`, which is a clock defined to represent simulation time. Therefore, it might be fragile to use their `count()` values like this, because it assumes that their tick period is the same.

Two options:
1. Insert `std::duration_cast` calls and cast them to a common type with the same tick period.
2. Insert a compile-time check like `static_assert(decltype(elapsedSimTime)::period == decltype(elapsedRealTime)::period);`

I think I prefer the first one if it works, but there may be arguments in favour of the latter too.

kyllingstad (changes requested):

src/cosim/timer.cpp
kyllingstad: I see now that this is just copied from old code elsewhere in the file. But then this is a good opportunity to fix it. ;)

include/cosim/timer.hpp
davidhjp01: I also  considered using `std::chrono::duration` but the `std::atmoic` documentation says:

> The primary std::atomic template may be instantiated with any [TriviallyCopyable](https://en.cppreference.com/w/cpp/named_req/TriviallyCopyable) type T satisfying both [CopyConstructible](https://en.cppreference.com/w/cpp/named_req/CopyConstructible) and [CopyAssignable](https://en.cppreference.com/w/cpp/named_req/CopyAssignable). The program is ill-formed if any of following values is false
    - [std::is_trivially_copyable](http://en.cppreference.com/w/cpp/types/is_trivially_copyable)<T>::value
    - [std::is_copy_constructible](http://en.cppreference.com/w/cpp/types/is_copy_constructible)<T>::value
    - [std::is_move_constructible](http://en.cppreference.com/w/cpp/types/is_move_constructible)<T>::value
    - [std::is_copy_assignable](http://en.cppreference.com/w/cpp/types/is_copy_assignable)<T>::value
    - [std::is_move_assignable](http://en.cppreference.com/w/cpp/types/is_move_assignable)<T>::value

Wasn't sure if this is the case for `std::chrono::duration` so just used `double` indicating seconds.

We can use mutex and create synchronized getter/setters after making the var private. Not sure how this would affect on the performance over `std::atomic` (since mutex locking/unlocking would be called on every step).

kyllingstad: I'm pretty sure `std::chrono::duration` is trivially copyable, so it should work.

conanfile.py
davidhjp01: I also think it should be okay with header only transitivity. But we had an issue in cmake finding the library without the option sometime ago. I will try to remove it and see if it is still causing the problem

src/cosim/timer.cpp
davidhjp01: Fixed.

include/cosim/timer.hpp
davidhjp01: Updated the comment

src/cosim/timer.cpp
davidhjp01: Added `duration_cast` as well as in `update_rolling_average_real_time_factor`.

davidhjp01: After adding duration_cast (or some other updates), I see noticeable performance drops. Investigating on it..

davidhjp01:
@kyllingstad  Two things I updated:
1. Using `std::is_same` to check if `cosim::time_point::period` and `Time::time_point::period` are the same type. If that is the case skipping `duration_cast`.
3. Using `int` for `sampling_period_to_monitor`.

I found that using `duration` type for `sampling_period_to_monitor` slows down the simulation quite a bit on my use-case (small step sizes) because the `duration` instance needs to be copied on every step for the hash computation.

Also additional `duration_cast` was quite an overhead. So trying to check the types are equal (e.g. nano) and if that is the case skipping `duration_cast` (this assumes the standard uses the same period type for the returning value when all `time_point`s in the expression also has the same `period` type?).

conanfile.py
davidhjp01: Seems okay after removal.

Issue #763, opened 2024-06-16
Clean up the test FMU set


kyllingstad:
The `tests/data/fmi2` directory contains a couple of FMUs from the old official "Test FMUs" set:

-`RoomHeating_OM_RH.fmu`
- `WaterTank_Control.fmu`

These are no longer available on the FMI web site, and as far as I can tell, they're no longer easily available online at all. We probably shouldn't be hosting obscure binary blobs in our repo.

And there are other issues too:

- There is no licence info for the abovementioned FMUs, which is bad form.
- Even our own FMUs are just binary blobs with no information about their origin or licence.
- `Clock.fmu` is our own,  but it's not part of [cpp-fmus](https://github.com/open-simulation-platform/cpp-fmus). Where is the source for that?

We really should clean this up. I would understand any user who is sceptical of running tests where we use these FMUs.

We do need to test with real FMUs, though, so any suggestions for improvements are welcome.

kyllingstad:
Actually, the `RoomHeating` FMU doesn't seem to be used by our tests, so it can simply be removed.

markaren:
They were generated by @yingguangchu. Ideally the SSP tests should just perform the same tests as the osp config, allowing us to remove these FMUs.

kyllingstad:
The OSP config tests use the same FMUs, it seems. But since we've now confirmed that the FMUs were contributed by NTNU to the OSP project, I'll assume that they are covered by the MPL 2.0 licence like most of the other things in the repo.

kyllingstad:
The most pressing issues were fixed in #764, so I'm closing this. Feel free to reopen if anyone sees the need to get rid of binary FMUs entirely.

PR #764, opened 2024-06-17
Clean up test FMUs

kyllingstad:
This fixes most of the issues described in #763. In particular, I have:

- removed FMUs with obscure origins
- removed unused FMUs (and some other unused test files)
- properly specified the origin and licence for the remaining FMUs
- modified tests so they only use the remaining FMUs

What I have not done is to banish pre-built FMUs entirely from the repo. That requires more discussion.


restenb (commented):
LGTM

PR #769, opened 2024-07-04
Enable exporting and importing subsimulator state

### LINKED ISSUE #756: Add interfaces to FMI functions for saving/serialising/deserialising/restoring FMU state

kyllingstad:
This is a follow-up to #765 and the second and final step to close #756. It is part of a PR series which will culminate in the implementation of #757, and until that point, I am targeting the `dev/state-persistence` branch rather than `master`.

Here, I've implemented functionality to export the internal state of individual subsimulators in a generic, structured form, and to import it again later.

This exported form is intended as an intermediate step before serialisation to disk.  The idea was to create a type that can be inspected and serialised to almost any file format we'd like.

The type is defined by `cosim::serialization::node` in `cosim/serialization.hpp`.  It is a hierarchical, dynamic data type with support for a variety of primitive scalar types and a few aggregate types: strings, arrays of nodes, dictionaries of nodes, and binary blobs. (Think JSON-like structure, only in-memory and with more types.)

Edit: `node` was originally a homebrew type. Now, it is based on Boost.PropertyTree. Otherwise, the description above still fits pretty well.


kyllingstad:
Hm. Seems like my little trick of storing an incomplete `node` type in an `unordered_map` doesn't work on GCC 9. It worked locally for me, but I use a more recent compiler. I've changed this PR to "draft" status while I figure out what to do about it.

kyllingstad:
Ok, I made it work now.

It seems that to make a recursive, map-like data type in standard C++, one has to hide the internal type, for example by messing about with `void*` pointers. Rather than do this myself, I have replaced the homemade `cosim::serialization::node` class with an alias for `boost::property_tree::basic_ptree`. The downside of this is a slightly less elegant API, but this is more than outweighed by the confidence we get from using tried-and-true code. Plus, we're already using `ptree` elsewhere in libcosim, so there's no extra dependency baggage involved.

include/cosim/slave.hpp
```diff
@@ -290,8 +291,6 @@ class slave
      *  `state_index`. The index is only valid for this particular slave.
      *
      *  The function may be called at any point after `setup()` has been called.
-     *
-     *  \pre `this->model_description().can_save_state`
```
kyllingstad, line 294: This and the removal on lines 304–305 was just something I missed in #765.

src/cosim/proxy/remote_slave.cpp
```diff
@@ -230,6 +230,21 @@ void cosim::proxy::remote_slave::release_state(state_index)
...
+}
+
+cosim::slave::state_index cosim::proxy::remote_slave::import_state(
+    const cosim::serialization::node&)
+{
+    throw error(
+        make_error_code(errc::unsupported_feature),
+        "Proxyfmu does not support state serialization");
+}
+
```
davidhjp01, line 247: I guess we do not support this until #757 gets finished?

src/cosim/serialization.cpp
```diff
@@ -0,0 +1,99 @@
...
+            return true;
+        }
+
+        bool operator()(std::byte value)
+        {
+            const auto flags = out_.flags();
+            out_ << "0x" << std::hex << int(value);
+            out_.flags(flags);
+            return true;
+        }
```
davidhjp01, line 48: This is not visited from `std::vector<std:;byte>`?

davidhjp01 (changes requested):
GW! One question - by looking at the visitor for byte type, are we going to export the FMU's serialized states into byte string?

tests/slave_simulator_unittest.cpp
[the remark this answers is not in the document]
davidhjp01: Or will this be implemented in #757 ?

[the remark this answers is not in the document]
kyllingstad: > Or will this be implemented in #757 ?

That was the idea. Note that I'm targeting this entire series on the `dev/state-persistence` side branch, so nothing will go into `master` before the entire pipeline has been implemented and tested. I just didn't want to put it all into one PR, as that would be a pain to review.

src/cosim/proxy/remote_slave.cpp
kyllingstad: That is something we should discuss, I guess. I just don't know the proxy-fmu code that well, and have enough on my plate with this, so I haven't prioritised it yet. That said, proxy-fmu is currently so important in our real-life use cases that I think it must be done at some point.

src/cosim/serialization.cpp
kyllingstad: First of all, note that this visitor is only used in `operator<<`, which is just intended as a tool for developing and debugging libcosim itself. I even suspect I'll remove it before this goes into `master`.

That said, the answer to your question is no. I was worried that some FMUs might have a lot of data in their exported state, so printing it all out as a hex sequence would be counterproductive. Also, the actual serialisation format is a hidden implementation detail of each FMU, so I'm not sure it would give us much.

For `std::vector<std::byte>` it therefore just prints `<binary data, N bytes>` where `N` is the size of the blob. I'm happy to reconsider this if you think the functionality would be generally useful for something.

src/cosim/proxy/remote_slave.cpp
restenb: We would need proxy-fmu support, but makes sense to leave that for last. Can be it's own PR following #757.

src/cosim/serialization.cpp
davidhjp01: Okay, I thought this would be used for actually serializing the entire states.

davidhjp01 (approved):

PR #770, opened 2024-07-09
Update cppreference tag file to 2019-06-07 version

kyllingstad:
This is the newest version available for download from https://en.cppreference.com/w/Cppreference:Archives at the time of writing. (In other words, I did not make these changes myself, so there's nothing to actually review.)

In case anyone is wondering, this is the file that enables us to hyperlink `std` symbols in our Doxygen-generated API documentation. I've updated it because some C++17 symbols are not included in our old tag file.


restenb (commented):
Approved by committee :)

PR #775, opened 2024-10-07
Controlling fixed precision for logging float point values

davidhjp01:
CSV files printed from `file_observer` does not include enough decimal places in case for logging latitude and longitude values for a ship vessel. Added a method `cosim::file_observer::file_observer_config::fixed_precision(const int)` to specify fixed precision value for `file_observer`.

### DIFF
--- a/src/cosim/observer/file_observer.cpp
+++ b/src/cosim/observer/file_observer.cpp
@@ -46,6 +46,7 @@ void clear_file_contents_if_exists(const cosim::filesystem::path& filePath, std:
         fsw.close();
     }
 }
+
 } // namespace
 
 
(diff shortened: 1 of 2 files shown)


src/cosim/observer/file_observer.cpp
```diff
@@ -301,10 +308,19 @@ class file_observer::slave_value_writer
...
+        const auto defaultPrecision = ss.precision();
+
         if (fsw_.is_open()) {
 
             for (const auto& [stepCount, times] : timeSamples_) {
-                ss << times << "," << stepCount;
+                if (precision_ > -1) {
+                    ss.precision(defaultPrecision);
+                    ss << times << "," << stepCount;
+                    ss.precision(precision_);
```
restenb, line 319: Maybe a way to avoid setting precision twice pr. sample is to use two string writers, one for custom & one for default precision. Then at the end should be able to just go `fsw_ << ss_default.rdbuf() << ss_custom.rdbuf(); `

restenb (approved):

src/cosim/observer/file_observer.cpp
davidhjp01: Each line will have time & stepCount so don't think we can concatenate at once after the loop.

Issue #771, opened 2024-07-18
Simplify API and code by dropping (alleged) support for changing system structure during simulation


kyllingstad:
In principle, the current libcosim API permits one to change the system structure after a simulation has started. More to the point, one is allowed to call `execution::add_simulator()`, `execution::connect_variables()`, etc. after `execution::simulate_until()` or `execution::step()` has been called. This was a deliberate design choice in the early OSP days, because we were told that the ability to dynamically add and remove entities was a desirable feature in certain human-in-the-loop simulation use cases.

Now, several years later, I am fairly certain that no one is using this for anything and that it doesn't work. Firstly, because we don't have tests for it, and secondly, for the following reason: If a subsimulator is added with `fixed_step_algorithm::add_simulator()` after `fixed_step_algorithm::initialize()` has been called, then the subsimulator’s initialisation functions (`simulator::setup()` and `simulator::do_iteration()`) will never be called. This, again, means that the underlying FMU will never have its `fmi2EnterInitializationMode()` and `fmi2ExitInitializationMode()` functions called, and then it’s not allowed to call its `fmi2DoStep()` function either.

**This leaves us with two choices: fix this feature or drop it. I say drop it**.

Dropping it would allow us to simplify the API in several places:

* An `execution` could be constructed directly from a `system_structure`, and we could remove `execution::add_simulator()`, `execution::add_function()`, and `execution::connect_variables()`.
* We could remove the homonymous functions plus `remove_simulator()`, and `disconnect_variable()` from the `algorithm` interface. (Note how those last functions aren't even exposed via `execution` yet, and that there is no `remove_function()`. It's all very half baked.)

But even more importantly, it would allow us to simplify the implementation:

*  The whole `inject_system_structure()` hack could be removed.
* `algorithm` implementations would be less complex and potentially more efficient, because their variable mapping tables are set up just once, they don't have to take into consideration that the variable mappings can change dynamically. 
* It would make [state saving](https://github.com/open-simulation-platform/libcosim/issues/768) and [persistence](https://github.com/open-simulation-platform/libcosim/issues/757) simpler, because one wouldn't have to keep data related to system structure at various levels (algorithm, execution), only at the outer level or perhaps not at all.

[...]

PR #777, opened 2024-11-06
Enable saving and restoring full simulation state

### LINKED ISSUE #768: Enable saving the entire simulation state in memory and restoring it again
I had originally intended this as a subtask of #757, which is about persisting the entire simulation state to disk, but after working on #756, I've found that it makes more sense to have it as a separate issue.

In essence, this issue is about implementing similar functionality as in #756, only on the full-system level. Here, we also need to export the state of the algorithm.

kyllingstad:
This closes #768. Some changes may warrant a bit of extra explanation:

**`execution` and `algorithm`:** I have taken one step towards prohibiting adding/removing sub-simulators after the co-simulation has begun, as decided in #771.  In particular, I've added the `execution::initialize()` function, which marks the point where such changes are no longer allowed.  (This is a backwards-compatible change, because it gets automatically called by `execution::step()` if it hasn't been called manually.)

**`slave_simulator`**: The changes in `slave_simulator.cpp` really ought to have been included in #769. Unfortunately, I didn't realise they were necessary before it was all put into the context of saving algorithm and execution state.  Basically, I thought I could get away with just saving each FMU's internal state, but it turns out that we also need to save the `slave_simulator` "get" and "set" caches.

(For those interested in the nitty-gritties, this is due to some subtleties regarding exactly when the cache values are set in the course of a co-simulation, relative to when the values are passed to the FMU. At the end of an "algorithm step", the "set cache" is out of sync with the FMUs input variables, and won't be synced before the next step. Simply performing an additional sync prior to saving the state is not sufficient, because that could have an effect on the FMUs output variables, thus invalidating the "get cache". That could in principle be updated too, but then the `slave_simulator` is in a whole different state from where it was when we started to save the state.)

### DIFF
--- a/src/cosim/execution.cpp
+++ b/src/cosim/execution.cpp
@@ -6,6 +6,7 @@
 #include "cosim/execution.hpp"
 
 #include "cosim/algorithm.hpp"
+#include "cosim/error.hpp"
 #include "cosim/exception.hpp"
 #include "cosim/slave_simulator.hpp"
 #include "cosim/utility/utility.hpp"
@@ -48,6 +49,7 @@ public:
         std::string_view name,
         duration stepSizeHint)
     {
+        COSIM_PRECONDITION(!initialized_);
         const auto index = static_cast<simulator_index>(simulators_.size());
         simulators_.push_back(std::make_unique<slave_simulator>(slave, name));
         algorithm_->add_simulator(index, simulators_.back().get(), stepSizeHint);
@@ -63,6 +65,7 @@ public:
 
     function_index add_function(std::shared_ptr<function> fun)
     {
+        COSIM_PRECONDITION(!initialized_);
         const auto index = static_cast<function_index>(functions_.size());
         functions_.push_back(fun);
         algorithm_->add_function(index, fun.get());
@@ -114,12 +117,7 @@ public:
         return currentTime_;
     }
 
-    bool is_running() const noexcept
-    {
-        return !stopped_;
-    }
-
-    duration step()
+    void initialize()
     {
         if (!initialized_) {
             algorithm_->initialize();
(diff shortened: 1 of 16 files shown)


src/cosim/slave_simulator.cpp
```diff
@@ -79,6 +79,38 @@ struct get_variable_cache
             }
         }
     }
+
+    serialization::node export_state() const
+    {
+        assert(references.size() == originalValues.size());
+        assert(std::all_of(modifiers.begin(), modifiers.end(), [](auto m) { return !m; }));
```
davidhjp01, line 86: Other option is to print warning if manipulators are currently applied and then continue exporting?

davidhjp01 (approved):
LGTM, just minor comments :)

davidhjp01:
Just leaving some ideas here - `nlohman_json` has export function to encode json data into CBOR. May be worth to consider switching it? Also I like its documentation :) https://json.nlohmann.me/api/basic_json/

src/cosim/execution.cpp
```diff
@@ -258,6 +267,52 @@ class execution::impl
...
+            throw;
+        } catch (...) {
+            throw error(
+                make_error_code(errc::bad_file),
+                "The serialized execution state is invalid or corrupt");
+        }
+
+        // Update observers
+        for (const auto& obs : observers_) {
+            obs->state_restored(lastStep_, currentTime_);
```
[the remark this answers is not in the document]
kyllingstad: Good point! Done.

src/cosim/slave_simulator.cpp
kyllingstad: Yeah, this could be a good option too. The main problem is that I am very unsure of what will happen if manipulators are active – under which circumstances it will work just fine, when it will subtly fail, and when it will fail spectacularly. I suspect that it will produce wrong results in the majority of cases, so I thought it was safer to make it an assertion rather than a warning. I think the whole “modifier” concept should be completely redesigned with a proper interface (à la `simulator`, `observer`, etc.) for this to work. It is just too easy to hide statefulness and aliasing behind a `std::function` object.

PR #780, opened 2024-12-10
Feature/779 state serialization

### LINKED ISSUE #779: Implement CBOR based de-/serializer for state import/export
Implementing de-/serialization methods for exporting and importing states to/from files

- [x] Serialization
- [x] Deserialization
- [x] Tests

davidhjp01:
closes #779 

- Uses [libcbor](https://github.com/PJK/libcbor) for (de)serialization.
- Overrides `<<` and `>>` to (de)seserialize `cosim::serialization::node` object to `std::o/istream`
- Added `state_serilization_test.cpp` which is based on `save_state_test.cpp` for testing. 
- Added `print_ptree` for replacing the previous `<<` operator for pretty-printing `cosim::serialization::node`.


davidhjp01:
@kyllingstad, @restenb the PR is ready for your reviews :)

davidhjp01:
I think we need to parse the following `CoSimulation` capabilities from each FMU's modelDescription file

- `canGetAndSetFMUstate`
- `canSerializeFMUstate`

so that users can check if they can export states from the FMUs they are running. I think FMUs not implementing these functions can throw exception or abruptly terminate?

restenb:
> I think we need to parse the following `CoSimulation` capabilities from each FMU's modelDescription file
> 
> * `canGetAndSetFMUstate`
> * `canSerializeFMUstate`
> 
> so that users can check if they can export states from the FMUs they are running. I think FMUs not implementing these functions can throw exception or abruptly terminate?

Not sure what the behavior would be, but it makes sense to include this check in any case. An FMU that does not set these flags should not be possible to use with state saving.

davidhjp01:
I found that the consumers of `libcosim` that tries to build `libcbor` during `conan install` process fails that saying it cannot find `libcbor.so` . I tried to add `transitive_libs=True, transitive_headers=True`, making `PUBLIC/PRIVATE` to cmake `target_link_libraries` but no luck.

I think this is bug in conan when upstream dependencies are added as shared library (related to  https://github.com/conan-io/conan/issues/13302) . Not sure if this is resolved in the newer version, but I am linking `libcobor` statically to avoid this issue. Any comments for better solutions are welcome @kyllingstad @restenb

src/cosim/observer/time_series_observer.cpp
```diff
@@ -263,10 +274,10 @@ void time_series_observer::simulator_step_complete(
 }
 
 void time_series_observer::state_restored(
-    step_number currentStep, time_point currentTime)
+    step_number, time_point)
 {
     for (const auto& entry : slaveObservers_) {
-        entry.second->observe(currentStep, currentTime);
+        entry.second->reset();
```
davidhjp01, line 280: I experienced an issue with time-series-observer when the state was restored.

The observer was not able to fetch correct time-series values via `get_step_numbers` and `get_real_samples` when the state was restored back in simulation time. This was due to the earlier simulation data in buffer `timeSamples_` was cleared (`adjustIfFull`). Because of this, the data needed in the restored state was not available anymore and the observer was no longer able to retrieve the correct time series data unless the simulation progresses until the time matches those in "max(`timeSamples_`)".

Since the data will be "re-generated" anyways after the restore, I suggest just to clear the buffers (both `timeSamples_` and `step_numbers`) to avoid this issue.

conanfile.py
```diff
@@ -39,19 +39,20 @@ def set_version(self):
...
+            self.requires("proxyfmu/0.3.3@osp/testing-feature_779-state-serialization",
+                          transitive_headers=True,
+                          transitive_libs=True)
         self.requires("yaml-cpp/[~0.8]")
         self.requires("xerces-c/[~3.2]")
 
     # Exports
     exports = "version.txt"
-    exports_sources = "*"
+    exports_sources = "*", "!cmake-build*", "!.venv", "!.git"
```
kyllingstad, line 55: Why is this change necessary? Do these files/dirs get created during the CI build process and end up in the packaged artifacts?

include/cosim/serialization.hpp
```diff
@@ -100,11 +100,16 @@ struct node_data_translator
 
 }} // namespace cosim::serialization
 
+/**
+ *  Serializes `data` into CBOR format and writes it to the output stream `out`.
+ */
+std::ostream& operator<<(std::ostream& out, const cosim::serialization::node& data);
```
kyllingstad, line 106: Something doesn't feel right about overloading `<<` and `>>` for this purpose. What if we want to add support for more serialisation formats? What if we want to add some additional settings to control how the serialisation is done?

src/cosim/fmi/v2/fmu.cpp
```diff
@@ -63,6 +63,12 @@ fmu::fmu(
     modelDescription_.description = fmi2_import_get_description(handle_);
     modelDescription_.author = fmi2_import_get_author(handle_);
     modelDescription_.version = fmi2_import_get_model_version(handle_);
+    modelDescription_.capabilities.can_get_and_set_fmu_state = !!fmi2_import_get_capability(
+        handle_,
+        fmi2_cs_canGetAndSetFMUstate);
+    modelDescription_.capabilities.can_serialize_fmu_state = !!fmi2_import_get_capability(
+        handle_,
+        fmi2_cs_canSerializeFMUstate);
```
kyllingstad, line 71: Note that these capabilities are already checked in `slave_instance::copy_current_state()` and `slave_instance::export_state()`, respectively, and exceptions are thrown on any attempt to save/serialise state of FMUs that don't support it.

Is there a specific use case that makes it more advantageous to read them upfront and store them in the `model_description`? Or, put differently, when and how would one use `model_description::capabilities`?

src/cosim/serialization.cpp
```diff
@@ -1,11 +1,22 @@
...
 
-#include <algorithm>
 #include <ios>
 #include <iterator>
 #include <utility>
+#include <cosim/log/logger.hpp>
+#include <iostream>
+#include <optional>
+#include <unordered_map>
+#include <cbor.h>
```
kyllingstad, line 10: These includes are not in the correct order, nor are they grouped appropriately. The correct order/format is:
```cpp
// First, main header for this `.cpp` file
#include "cosim/serialization.hpp"

// Then other libcosim headers (using double quotes in library-internal `.cpp` files)
#include "cosim/log/logger.hpp"

// Then third-party libraries
#include <cbor.h>

// Then stdlib headers
#include <ios>
#include <iostream>
#include <iterator>
#include <optional>
#include <unordered_map>
#include <utility>
```
Alphabetical order should be used within groups.

src/cosim/serialization.cpp
```diff
@@ -1,11 +1,22 @@
...
 
 
+namespace cosim
+{
+namespace serialization
+{
+} // namespace serilization
+} // namespace cosim
+
```
kyllingstad, line 19: Why the empty namespaces?

src/cosim/serialization.cpp
```diff
@@ -75,25 +86,507 @@ namespace
...
 
+    constexpr uint64_t TYPE_INT  = 0x8000;
+    constexpr uint64_t TYPE_UINT = 0x8001;
+
+    template<typename F, typename... Args>
+    void wrap_cbor_call(F&& f, Args&&... args)
+    {
+        auto result = std::forward<F>(f)(std::forward<Args>(args)...);
+        if (!result) {
+            throw std::runtime_error("Failed Cbor operation!");
```
kyllingstad, line 111: Is there any way to get more info about the error and include it in the exception? It would be very useful for the user to know e.g. whether it failed because of a wrong path, insufficient permissions, etc.

If libcbor uses standard `errno` error codes, you may be able to use [`system_error`](https://en.cppreference.com/w/cpp/error/system_error/system_error) and its related functionality for this.

src/cosim/serialization.cpp
```diff
@@ -75,25 +86,507 @@ namespace
...
+    in.clear();
+    in.seekg(0, std::ios::beg);
+    std::vector<uint8_t> buf(length);
+    in.read(reinterpret_cast<char*>(buf.data()), length);
+
+    // Initializing the reader context object
+    cbor_reader_ctx ctx{root};
+
+    struct cbor_callbacks cbs = cbor_empty_callbacks;
+    struct cbor_decoder_result decode_result{};
```
kyllingstad, line 549: You don't need to prefix these with `struct` in C++, that's a C thing.

src/cosim/serialization.cpp
```diff
@@ -75,25 +86,507 @@ namespace
...
+    cbs.null = cbor_reader::cbor_read_null_undefined;
+    cbs.tag = cbor_reader::cbor_read_tag;
+    cbs.byte_string = cbor_reader::cbor_read_byte_string;
+    cbs.indef_array_start = cbor_reader::cbor_indef_array_start;
+    cbs.array_start = cbor_reader::cbor_array_start;
+
+    while(bytes_read < len) {
+        decode_result = cbor_stream_decode(buf.data() + bytes_read, len - bytes_read, &cbs, &ctx);
+        if (decode_result.status != cbor_decoder_status::CBOR_DECODER_FINISHED) {
+            BOOST_LOG_SEV(cosim::log::logger(), cosim::log::error) << "[Cbor reader] Decoding error " << decode_result.status;
```
kyllingstad, line 579: Why not throw an exception here? Won't this mean that the resulting object is incomplete, and therefore not useful for restoring a simulation state?

tests/CMakeLists.txt
```diff
@@ -49,7 +50,7 @@ foreach(test IN LISTS tests)
         "cpp_${test}"
         FOLDER "C++ tests"
         SOURCES "${test}.cpp"
-        DEPENDENCIES cosim
+        DEPENDENCIES "cosim;libcbor::libcbor"
```
kyllingstad, line 53: Why does libcbor need to be listed here when none of our other dependencies are?

tests/data/fmi2/README.md
```diff
@@ -1,11 +1,12 @@
 # FMI 2.0 test FMUs
 
 | Name                   | Origin                  | License                                  |
-| ---------------------- | ----------------------- | ---------------------------------------- |
+|------------------------|-------------------------|------------------------------------------|
 | `CraneController.fmu`  | OSP                     | [MPL 2.0](../../../LICENSE)              |
 | `Dahlquist.fmu`        | [Reference FMUs] 0.0.31 | [2-clause BSD](./reference-fmus_LICENSE) |
 | `KnuckleBoomCrane.fmu` | OSP                     | [MPL 2.0](../../../LICENSE)              |
 | `vector.fmu`           | [OSP cpp-fmus]          | [MIT](./osp_cpp-fmus_LICENSE)            |
+| `BouncingBall.fmu`     | [Reference-FMUs] 0.0.37 | [2-clause BSD](./BouncingBall_LICENSE)   |
```
kyllingstad, line 9: If this is one of the reference FMUs, you could just reuse the link (and file) on line 6.

tests/save_state_test.cpp
```diff
@@ -1,15 +1,20 @@
...
+#include <boost/property_tree/json_parser.hpp>
+#include <cosim/serialization.hpp>
 
 #include <exception>
 #include <limits>
 #include <memory>
 #include <stdexcept>
 #include <iostream>
+#include <chrono>
+#include <sstream>
```
kyllingstad, line 17: Include order and grouping should be fixed.

kyllingstad (changes requested):
This is great, thanks for doing this! Overall, I think it looks good, and most of my comments are minor nitpicks. The main issues I have, on which I'd like some justification/discussion, are:

1. the overloading of `<<` and `>>` for serialisation/deserialisation
2. the need to preload the FMU capabilities and expose them in `model_description`

I've added specific comments on each of these.

conanfile.py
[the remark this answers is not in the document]
kyllingstad: In that case, I'd suggest specifying the files and directories we want to export, rather than excluding the ones we *don't* want to export, which is going to vary greatly depending on the individual developer and their toolchain. (For example, in my workdir, there is no `.venv/` or `cmake-build/`, but instead I have `build/`, `compile_commands.json`, and more.) The list of files to include is much more predictable, and is not likely to change often.

conanfile.py
```diff
@@ -66,6 +67,7 @@ def configure(self):
         if self.options.shared:
             self.options.rm_safe("fPIC")
         self.options["*"].shared = self.options.shared
+        self.options["libcbor/*"].shared = False
```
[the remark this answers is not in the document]
davidhjp01: I made libcbor shared library and also declared as "PUBLIC" when linking in CMake. 

Linux consumers may still require to declare `libcbor` in their recipe to avoid an error during CMake configuration stage. If this is not preferable I think we need to link it statically until conan addresses this issue

src/cosim/fmi/v2/fmu.cpp
davidhjp01: I thought it would be beneficial to have this information to the user before running the fmu instead of try and catch exception by actually running those functions. In this way we can inform users that the fmu(s) he/she is simulating support this capability and also help consumers of libcosim act accordingly.

conanfile.py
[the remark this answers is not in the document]
davidhjp01: Updated

tests/save_state_test.cpp
davidhjp01: Fixed

tests/data/fmi2/README.md
davidhjp01: Updated tests to use Dahlquist fmu instead

tests/CMakeLists.txt
davidhjp01: Removed, not needed after update.

src/cosim/serialization.cpp
davidhjp01: Fixed

src/cosim/serialization.cpp
davidhjp01: Not sure what happened here, removed.

src/cosim/serialization.cpp
davidhjp01: Fixed. But it seems `logger.cpp/hpp` is a public header/source.

include/cosim/observer/file_observer.hpp
```diff
@@ -88,9 +89,21 @@ class file_observer_config
      */
     static file_observer_config parse(const filesystem::path& configPath);
 
+    /**
+     * Sets fixed precision value for floating point numbers
+     *
+     * \param precision the number of digits after the decimal point
```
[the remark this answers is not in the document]
davidhjp01: Updated

include/cosim/model_description.hpp
```diff
@@ -195,6 +202,9 @@ struct model_description
 
     /// Variable descriptions.
     std::vector<variable_description> variables;
+
+    /// FMI 2.0 Co-simulation capabilities
```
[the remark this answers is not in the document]
davidhjp01: I am not sure, would it be okay then at least not mentioning "FMI" in the comment?

conanfile.py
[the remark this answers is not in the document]
kyllingstad: I think this is ok. It's easy enough for consumers to specify `-o libcbor:shared=False` on the Conan command line too if they prefer static linking.

P.S. I noticed now that you had [commented](https://github.com/open-simulation-platform/libcosim/pull/780#issuecomment-2619514686) on this issue long before I submitted my review. Sorry for not noticing before.

include/cosim/model_description.hpp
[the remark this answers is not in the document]
kyllingstad: I think "model capabilities" or "simulator capabilities" should be sufficient.

src/cosim/fmi/v2/fmu.cpp
kyllingstad: Ok, fair enough. Then I suggest rewriting the existing checks in `copy_current_state()` and `export_state()` so they use the new `model_description::capabilities` flags rather than re-calling the fmilib functions.

src/cosim/serialization.cpp
kyllingstad: Yeah, and this is something you'll see people having different opinions about. But the rule we've used thus far is to always use `#include <cosim/...>` in public headers (i.e., anything under `include/`), since those should *only* depend on other public headers, and to always use `#include "cosim/..."` in `.cpp` files, since they *may* pull in private headers.

One *could* imagine using angle brackets when including public headers and quotes when including private headers in `.cpp` files, but that would be too much of a hassle. Most of us probably don't remember which is which and would have to check every time. It's easier this way.

include/cosim/serialization.hpp
davidhjp01: Update the code so that we can now do something like: 
```cpp
os << cosim::serialization::format::cbor << data
```
So that another format can be added if needed.

src/cosim/serialization.cpp
davidhjp01: Hmm unfortunately I do not see in the document that we can retrieve an error code object if the functions fail (for encoding).

davidhjp01: I am now passing an err message to the function so that at least we know at which stage encoding has failed.

src/cosim/fmi/v2/fmu.cpp
davidhjp01: Updated

include/cosim/model_description.hpp
```diff
@@ -59,8 +59,8 @@ enum variable_variability
     continuous
 };
 
-/// Subset of Co-Simulation v2 capabilities in the XML schema
-struct cosim_capabilities
+/// A list of simulator capabilities
+struct simulator_capabilities
 {
     bool can_get_and_set_fmu_state{};
```
kyllingstad, line 65: Sorry, I missed this in my previous review, but these should not have `fmu` in their names, for the same reasons as mentioned earlier. Maybe just `can_save_state` and `can_export_state`, i.e., using the same terminology as the corresponding API functions?

kyllingstad (approved):
Looks mostly good now, thanks for implementing the requested changes. I have one minor change request still, but I'll approve so you can just merge after fixing it.

include/cosim/model_description.hpp
kyllingstad: Btw, this is a stylistic preference, but I think it is clearer to be explicit about the initialisation of primitive variables like these, and use `= false` rather than `{}`.

davidhjp01: Updated as suggested. 

I will merge this PR after merging the proxyfmu PR.

kyllingstad:
The question now is what do we do with the `dev/state-persistence` branch? Do we think it is ready to merge into `master`?

davidhjp01:
> The question now is what do we do with the `dev/state-persistence` branch? Do we think it is ready to merge into `master`?

I think so, but maybe we just need to quick look through it and test before merge (we will test the branch on our dev environment for a bit, just in case 😊).  This PR was a fork of `dev/state-persistence' so we already have fixed some of the issues and tested by closing this PR.

PR #783, opened 2025-02-07
Feature: ECCO Algorithm

restenb:
At long last, a working implementation of ECCO that also supports configuring powerbonds from the system structure. 

Some choices had to be made to allow this, but here are the most relevant changes:

* `<Algorithm>` in OspSystemStructure now accepts either `ecco` or `fixedStep`.
* The configuration for the Ecco algorithm can be added to the root of `OspSystemStructure `as follows:

```
    <EccoConfiguration>
        <SafetyFactor>0.99</SafetyFactor>
        <StepSize>0.0001</StepSize>
        <MinimumStepSize>0.00001</MinimumStepSize>
        <MaximumStepSize>0.01</MaximumStepSize>
        <MinimumChangeRate>0.2</MinimumChangeRate>
        <MaximumChangeRate>1.5</MaximumChangeRate>
        <ProportionalGain>0.2</ProportionalGain>
        <IntegralGain>0.15</IntegralGain>   
        <RelativeTolerance>1e-6</RelativeTolerance>
        <AbsoluteTolerance>1e-6</AbsoluteTolerance>
    </EccoConfiguration>
```
* To describe a powerbond, add the attribute `powerbond="mypowerbond"` to either a `VariableConnection `or `VariableGroupConnection` element. This defines a name for the powerbond that the parsing uses to group correctly.
* To define the individual variables, the attribute `port` has been added to the `Variable` element. Here the user must specify whether the variable is the `input` or `output` port of it's side of the bond. So, if we are coupling for instance a force <-> velocity bond, this results in a tuple with one input and one output port for each `VariableConnection` that is used in the bond. 
* This information is then parsed by `osp_config_parser`, which results in a `power_bond_map `available through the `system_structure` object.
* Finally, the `power_bond_map` is iterated and power bonds added to the algorithm by `execution::inject_system_structure`. The test `ecco_algorithm_from_system_structure_test.cpp` has been added to showcase this.

This PR is on the larger side, and while there aren't any breaking changes to worry about, input on improvements to the code and usability are very welcome.


include/cosim/system_structure.hpp
```diff
@@ -213,8 +213,47 @@ class system_structure
...
+        full_variable_name u_a{"", ""};
+
+        /// The u_b variable in the bond, according to the description of the ecco algorithm.
+        full_variable_name u_b{"", ""};
+
+        /// The y_a variable in the bond, according to the description of the ecco algorithm.
+        full_variable_name y_a{"", ""};
+
+        /// The y_b variable in the bond, according to the description of the ecco algorithm.
+        full_variable_name y_b{"", ""};
```
kyllingstad, line 229: These variable names are not very descriptive. I suggest replacing `u` with `input` and `y` with `output`, and possibly swapping the order of the second and third declaration.

include/cosim/system_structure.hpp
```diff
@@ -251,6 +290,9 @@ class system_structure
         add_entity({std::string(name), type, {}, std::move(parameters)});
     }
 
+    power_bond_map get_power_bonds() const noexcept;
+    void add_power_bond(std::string name, power_bond pb);
```
kyllingstad, line 294: These should be documented.

src/CMakeLists.txt
```diff
@@ -65,6 +66,7 @@ set(privateHeaders
 )
 set(sources
     "cosim/algorithm/fixed_step_algorithm.cpp"
+    "cosim/algorithm/ecco_algorithm.cpp"
```
kyllingstad, line 69: These should be in alphabetical order too.

src/cosim/algorithm/ecco_algorithm.cpp
```diff
@@ -0,0 +1,646 @@
...
+#include <unordered_map>
+#include <vector>
+
+
+namespace cosim
+{
+namespace
+{
+
+int calculate_decimation_factor(
```
kyllingstad, line 28: Here is an example of something that we can get rid of by dropping multi-rate support.

src/cosim/execution.cpp
```diff
@@ -12,6 +12,7 @@
 
 #include <algorithm>
 #include <atomic>
+#include <iostream>
```
kyllingstad, line 15: Leftover `#include` from a debugging session?

src/cosim/execution.cpp
```diff
@@ -549,6 +560,22 @@ entity_index_maps inject_system_structure(
             val);
     }
 
+    // For ECCO algorithm, add the powerbonds
+    if (sys.algorithm == "ecco") {
+        const auto powerbonds = sys.get_power_bonds();
+        auto algo = (cosim::algorithm*)exe.get_algorithm().get();
+        auto algorithm = reinterpret_cast<cosim::ecco_algorithm*>(algo); /// ugly downcasting, better alternatives?
```
kyllingstad, line 567: Yes, use [`std::dynamic_pointer_cast()`](https://en.cppreference.com/w/cpp/memory/shared_ptr/pointer_cast). In light of my request to remove `system_structure::algorithm` again, however, I suggest a change to the entire `if` block – something like this:
```cpp
// For ECCO algorithm, add the power bonds
if (auto algorithm = std::dynamic_pointer_cast<ecco_algorithm>(exe.get_algorithm())) {
    const auto powerBonds = sys.get_power_bonds();
    if (powerBonds.empty()) {
        throw /* some exception saying that ECCO won't work without power bonds */;
    }
    // ...
```

include/cosim/system_structure.hpp
```diff
@@ -213,8 +213,47 @@ class system_structure
...
+            y_a = ya;
+        }
+
+        void set_yb(full_variable_name& yb)
+        {
+            y_b = yb;
+        }
+    };
+
+    std::string algorithm;
```
kyllingstad, line 252: `system_structure` is only supposed to describe the modelled *system*; I don't think it should contain knowledge/requirements concerning the algorithm. (It is fine to have the info about the power bonds here, though, just document that not all algorithms will use it.)

include/cosim/algorithm/ecco_algorithm.hpp
```diff
@@ -0,0 +1,103 @@
...
+ */
+#ifndef LIBCOSIM_ALGORITHM_ECCO_ALGORITHM_HPP
+#define LIBCOSIM_ALGORITHM_ECCO_ALGORITHM_HPP
+
+#include <cosim/algorithm/algorithm.hpp>
+
+namespace cosim
+{
+
+struct ecco_parameters
```
kyllingstad, line 18: See comment about naming wrt. `fixed_step_configuration`.

include/cosim/osp_config_parser.hpp
```diff
@@ -24,6 +25,9 @@ struct osp_config
     /// The system structure
     cosim::system_structure system_structure;
 
+    // The algorithm
+    std::variant<cosim::fixed_step_configuration, cosim::ecco_parameters> algorithm_configuration;
```
kyllingstad, line 29: Now we have this, can we remove the `step_size` field further down?

tests/CMakeLists.txt
```diff
@@ -1,5 +1,8 @@
 set(tests
     "fixed_step_algorithm_test"
+    "ecco_algorithm_test"
```
kyllingstad, line 3: Alphabetical order.

tests/ecco_algorithm_multi_bond_test.cpp
kyllingstad, line 1: Please remove commented-out lines in this file.

tests/data/ecco/quarter_truck/Mass.fmu
kyllingstad, line 1: FMUs should be in either `tests/data/fmi1/` or `tests/data/fmi2/` depending on FMI version. Furthermore, the README file in that directory should be updated with information about origin, and the FMUs' licence(s) should be included in the same directory. (See #763 for background.)

kyllingstad (changes requested):
Awesome, great to finally get this done! I have quite a few comments, but they mostly concern details. The overall approach seems good!

conanfile.py
```diff
@@ -51,7 +51,7 @@ def requirements(self):
 
     # Exports
     exports = "version.txt"
-    exports_sources = "*"
+    exports_sources = "*", "!cmake-build*", "!.venv", "!.git"
```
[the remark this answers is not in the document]
kyllingstad: Btw, @davidhjp01 already fixed it in #780.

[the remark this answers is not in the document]
restenb: Yep, intention was to sync with @davidhjp01's changes here.

include/cosim/system_structure.hpp
restenb: Good idea, `u` and `y` were taken from the paper of course :). Replaced `u` with `input` and `y` with `output` everywhere.

tests/data/ecco/quarter_truck/Mass.fmu
restenb: Moved the `quarter_truck `folder to` /fmi2` and deleted the `ecco `folder. Deleted `spring`, `mass`, and `damper ` FMUs since they are not used in the ECCO tests. Added LICENSE (MIT in this case) from the quarter-truck in osp-demo-cases which is where the FMUs were taken from.

restenb:
> Awesome, great to finally get this done! I have quite a few comments, but they mostly concern details. The overall approach seems good!

Thanks! Detailed feedback is always appreciated. :) I've gone over & addressed most comments now. 

To properly "tie a bow on it" we need to resolve which models are to be used for the ECCO tests. The quarter truck case (chassis, wheel, ground FMUs) seems appropriate since it immediately lets you define the power bonds between the models, but I discovered that the committed ones are slightly different from the models in OSP demo-cases. This is likely because we wanted some more control over model internals and OS for testing, but I do not remember exactly where we got this version of `wheel` and `chassis`. We are also missing a corresponding version of `ground` to extend the tests. 

All the linux builds are failing now due to common test failures:

`[error  ] [FMI Library: FMICAPI] Could not load the FMU binary: /lib/x86_64-linux-gnu/libm.so.6: version `GLIBC_2.29' not found (required by /tmp/libcosim_36760552-a42c-4210-aedb-9d98732b3b57/e3805d85-0e10-3f84-b688-4612d91069e8/binaries/linux64/Wheel.so)`

This is likely because these model were compiled on a newer version of linux than the `gcc9:ubuntu-16.04` image we use to build libcosim. So if we go with these models, they'll have to be recompiled from source on ubuntu16 also I assume, which adds some work.

davidhjp01:
Just one comment, once we have the updated wheel model, should we just clone/download from the case-study repository rather than committing it to libcosim?

SSkjong:
Those models could be from fmiCpp, I know David and I made some test models a few years ago.

restenb:
> Just one comment, once we have the updated wheel model, should we just clone or download the fmu file from the case-study repository rather than committing it to libcosim?

That is also an option yes, to update the demo-cases FMUs with a linux binary as well for example. 

I'm not sure if it'd still be easier to keep a local copy of the models for end users though, rather than rely on the build system to download them.

restenb:
> Those models could be from fmiCpp, I know David and I made some test models a few years ago.

I'd say almost certainly they are - question is, where is the source? 😊

SSkjong:
> > Those models could be from fmiCpp, I know David and I made some test models a few years ago.
> 
> 
> 
> I'd say almost certainly they are - question is, where is the source? 😊

I will share it with you.

include/cosim/algorithm/ecco_algorithm.hpp
```diff
@@ -0,0 +1,103 @@
...
+    double min_change_rate;
+    double max_change_rate;
+    double abs_tolerance;
+    double rel_tolerance;
+    double p_gain;
+    double i_gain;
+};
+
+/**
+ *  A fixed-stepsize co-simulation algorithm.
```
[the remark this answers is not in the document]
restenb: Updated. Note: Decided to namedrop the original ECCO paper as the source also.

include/cosim/algorithm/ecco_algorithm.hpp
```diff
@@ -0,0 +1,103 @@
...
+     * of the algorithm's base step size. The default decimation factor is 1.
+     * Must be called *after* the simulator has been added to the algorithm with `add_simulator()`.
+     *
+     * \param simulator
+     *      The index of the simulator.
+     *
+     * \param factor
+     *      The stepsize decimation factor.
+     */
+    void set_stepsize_decimation_factor(simulator_index simulator, int factor);
```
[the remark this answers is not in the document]
restenb: Removed `set_stepsize_decimation_factor` as suggested.

src/cosim/algorithm/ecco_algorithm.cpp
restenb: Dropped this. 

I don't recall why we did it - but I see that also e.g. transferring variables is only done when the least common multiple of the source and target simulator matches:

```
            const auto sdf = simulators_.at(c.source.simulator).decimationFactor;
            const auto tdf = functions_.at(c.target.function).decimationFactor;
            if (stepCounter_ % std::lcm(sdf, tdf) == 0) {
                transfer_variable(c);
            }
```

Should we remove this as well, and instead throw an error if the simulators have different rates? Or am I misunderstanding something?

include/cosim/system_structure.hpp
restenb: Added it reflexively, but it wasn't actually used anywhere in the code yet. So removed.

include/cosim/osp_config_parser.hpp
restenb: Yes. :D

include/cosim/system_structure.hpp
restenb: I was more thinking of `system_structure `as a representation of the `OspSystemStructure`. 

If we don't add it here, we'll need to implement a different way of exposing which algorithm is defined in the XML.

src/cosim/algorithm/ecco_algorithm.cpp
kyllingstad: Yes, I think the `decimationFactor` variables and everything else having to do with multi-rate co-simulation should be entirely removed from this file. It's an algorithm-internal thing anyway, not part of the `algorithm` interface in any way.

What *is* part of the interface is the `stepSizeHint` parameter of the `cosim::algorithm::add_simulator()` function. But as the name implies, it is just a hint, and the idea is that the algorithm should make a “best effort” to respect it. (In `fixed_step_algorithm` we support multi-rate, and there I believe we round the step size hint to the nearest multiple of the base synchronisation rate.)

An interesting question is whether we could/should use `stepSizeHint` for something in the ECCO implementation. One idea could be to gather the hints for all subsimulators and use the smallest one as the initial step size. Calling Dr. @ssadjina for advice here.

restenb: The rest of the decimation factor code has also been removed from `ecco_algorithm`.

restenb:
I have now recompiled the fmicpp models on Ubuntu1604, and the build is going again. 

Did not find the source for `ground` there, so that has been skipped for now. 

Hopefully we can merge this during the upcoming committee meeting. :)

tests/data/fmi2/quarter_truck/README.md
```diff
@@ -0,0 +1,10 @@
+# FMI 2.0 test FMUs
+
+| Name                   | Origin         | License                       |
+| ---------------------- | -------------- | ----------------------------- |
+| `Chassis.fmu`  | OSP            | [MIT](LICENSE)   |
+| `Wheel.fmu` | OSP            | [MIT](LICENSE)   |
+| `Ground.fmu`           | [MIT] | [MIT](LICENSE) |
```
kyllingstad, line 7: The spacing here is weird. Also, cell 3:2 should probably not contain `[MIT]`.

restenb:
Further discussion points from committee meeting:

* The use of `port` for power bonds is ambiguous since it is also used in OSP-IS. Suggestion to use `causality` instead for now.
*  Ideally, using the OSP-IS connection types makes the configuring of power bonds easier since e.g. LinearMechanicalPort already describes a bond. Supporting this would require some further work in the already congested `osp_config_parser.cpp ` however, but we do still want to support OSP-IS.
* At the same time, having the option to configure on the "lowest" level directly on the variables is probably needed, but it might be made simpler by deducing signal directions from the `ModelDescription`.
* In the same vein, there is a question of whether declaring `input `and `output `ports are desired, or even necessary, if we're able to deduce the bond and the correct signal directions from the `OSPSystemStructure `and `OSPModelDescription `we should aim to do so.

My suggestion is to change port to causality for now, and clean up a couple smaller issues like leftover comments that were seen, then merge so we can start using ECCO. New tickets will be created for suggested further improvements of the ECCO implementation.

tests/data/fmi2/quarter_truck/README.md
restenb: Cleaned this up & removed ground. LICENSE was taken from the FMICPP repo, which appears to be MIT.

kyllingstad (approved):
Looks good now!

I am still not super happy about the fact that one has to redundantly specify the causality of variables in `OspSystemStructure.xml` when this information is already present in the model description, but I accept that this PR is already complicated enough and we can leave it for later. I suggest you create an issue about it, though, so it's not forgotten.

data/xsd/OspSystemStructure.xsd
```diff
@@ -16,6 +16,7 @@
                 <xs:element name="Simulators" type="osp:simulators"/>
                 <xs:element name="Functions" type="osp:functions" minOccurs="0"/>
                 <xs:element name="Connections" type="osp:connections" minOccurs="0"/>
+                <xs:element name="EccoConfiguration" type="osp:eccoconfiguration" minOccurs="0"/>
```
davidhjp01, line 19: need to specify maxOccurs?

include/cosim/algorithm/ecco_algorithm.hpp
```diff
@@ -0,0 +1,97 @@
...
+    /**
+     *  Constructor.
+     *
+     *  \param baseStepSize
+     *      The base communication interval length.
+     *
+     *  \param workerThreadCount
+     *      The number of worker threads to spawn for running FMUs
+     */
+    explicit ecco_algorithm(ecco_algorithm_params params, std::optional<unsigned int> workerThreadCount = std::nullopt);
```
davidhjp01, line 51: FYI, I quickly tried but now i could pass variant type without an error

include/cosim/system_structure.hpp
```diff
@@ -213,8 +213,25 @@ class system_structure
...
+        full_variable_name input_a{"", ""};
+
+        /// The output_a variable in the bond, according to the description of the ecco algorithm.
+        full_variable_name output_a{"", ""};
+
+        /// The input_b variable in the bond, according to the description of the ecco algorithm.
+        full_variable_name input_b{"", ""};
+
+        /// The output_b variable in the bond, according to the description of the ecco algorithm.
+        full_variable_name output_b{"", ""};
```
davidhjp01, line 229: nitpicking, but I would add a default ctor to avoid passing the empty strings

src/cosim/algorithm/ecco_algorithm.cpp
```diff
@@ -0,0 +1,600 @@
...
+#include <numeric>
+#include <sstream>
+#include <unordered_map>
+#include <vector>
+
+
+namespace cosim
+{
+
+class ecco_algorithm::impl
```
davidhjp01, line 26: Not need to be addressed in this PR, but I think a lot of methods here are duplicates from `fixed_step_algorithm`. Not sure if it would cost more compilation time but we can create a base class for the algorithm `impl` for reusing those methods.

tests/ecco_algorithm_from_system_structure_test.cpp
davidhjp01, line 1: Remove std outs in tests?

tests/ecco_algorithm_multi_bond_test.cpp
davidhjp01, line 1: Same here (removing std outs)

tests/ecco_algorithm_test.cpp
davidhjp01, line 1: To remove std outs

tests/mock_slave.hpp
```diff
@@ -35,11 +35,11 @@ class mock_slave : public cosim::slave
...
-        std::function<double(cosim::time_point, double)> realOp = nullptr,
-        std::function<int(cosim::time_point, int)> intOp = nullptr,
-        std::function<bool(cosim::time_point, bool)> boolOp = nullptr,
-        std::function<std::string(cosim::time_point, std::string_view)> stringOp = nullptr,
-        std::function<void(cosim::time_point)> stepAction = nullptr)
+        std::function<double(cosim::time_point, cosim::duration, double)> realOp = nullptr,
+        std::function<int(cosim::time_point, cosim::duration, int)> intOp = nullptr,
+        std::function<bool(cosim::time_point, cosim::duration, bool)> boolOp = nullptr,
+        std::function<std::string(cosim::time_point, cosim::duration, std::string_view)> stringOp = nullptr,
+        std::function<void(cosim::time_point, cosim::duration)> stepAction = nullptr)
```
davidhjp01, line 42: I remember we added this `cosim::duration` previously for ecco. But I think then these functions do not really emulate `do_step` in FMI spec, would this be conceptually okay?

One way to avoid this is to store the previous time_points in the test script (that uses mock_slave) and pass them as refs and compute duration within lambda.

davidhjp01 (approved):
Great work Roger! I two main comments/questions:

1. If the osp_config_parser checks power bond name uniqueness?
2. For `mock_slave` should we keep passing step duration or remove it (see the comments above)

include/cosim/system_structure.hpp
kyllingstad: Ah, nice catch and good point. `std::string("")` is not the same as `std::string()`!

src/cosim/algorithm/ecco_algorithm.cpp
kyllingstad: There may be a potential to avoid duplication, but I don't think that inheritance is the way to exploit it, and I don't think we need to worry about it quite yet. There are two rules of thumb I like to follow in cases like this:

* Prefer [composition over inheritance](https://en.wikipedia.org/wiki/Composition_over_inheritance)
* The [rule of three](https://en.wikipedia.org/wiki/Rule_of_three_(computer_programming))

data/xsd/OspSystemStructure.xsd
restenb: Good idea. In fact, it might be a good idea to also specify `maxOccurs` for `Algorithm`, `BaseStepSize`, `StartTime `and `EndTime`. So I did just that.

include/cosim/algorithm/ecco_algorithm.hpp
restenb: Pass `variant` to where? This constructor? 

TBH I'm not sure if/when variant types can be deduced by the compiler, or if it is always necessary to `std::visit` or `std::get` the specific type.

restenb:
> I am still not super happy about the fact that one has to redundantly specify the causality of variables in `OspSystemStructure.xml` when this information is already present in the model description, but I accept that this PR is already complicated enough and we can leave it for later. I suggest you create an issue about it, though, so it's not forgotten.

I agree, and I think this is natural follow-on work for increased convenience when ECCO starts seeing use. Created https://github.com/open-simulation-platform/libcosim/issues/786 to track this.

src/cosim/osp_config_parser.cpp
```diff
@@ -682,6 +788,109 @@ osp_config_parser::SignalConnection find_signal_connection(
...
+        std::vector<osp_config_parser::PowerBondConnection> aPowerBond;
+        std::copy_if(pbConnections.begin(), pbConnections.end(), std::back_inserter(aPowerBond), [pbName](osp_config_parser::PowerBondConnection conn) {
+            return pbName == conn.name;
+        });
+
+        if (aPowerBond.size() != 2) {
+            std::ostringstream oss;
+            oss << "The powerbond " << pbName << " does not have 2 connections! Number of connections found was " << aPowerBond.size();
+            throw std::runtime_error(oss.str());
+        }
```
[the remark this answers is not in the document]
restenb: That would be a reasonable check to have, indeed. Implemented now by adding all power bond names to a vector, sorting it, and calling `std::unique` to give a count of unique names (each unique name should be found twice, which is implicitly checked by ensuring that there are two connections in each bond). The number of unique names should then be equal to the number of power bonds in the system structure.

include/cosim/algorithm/ecco_algorithm.hpp
davidhjp01: Yep, but I think I am fine to continue with the current version.

include/cosim/system_structure.hpp
restenb: Not sure what you're getting at exactly. Add a default ctor for `full_variable_name`? 

The current ctor just invokes std::move on each string:

```
    full_variable_name(std::string simulatorName, std::string variableName)
        : entity_name(std::move(simulatorName))
        , variable_name(std::move(variableName))
    { }
```

tests/mock_slave.hpp
restenb: I guess having the option to export & import the current step size would be required eventually, if we ever want to combine ECCO with state saving.

restenb:
Further comments have been followed up now, and a couple issues that occurred due to merging in the state saving PR from master have been fixed. Will merge shortly.

include/cosim/system_structure.hpp
davidhjp01: I think you can make a default ctor and initialize those strings in the declaration or create a ctor without argument and initialize them in there.

kyllingstad: The easiest is just to create a default constructor like this:
```
full_variable_name() = default;
```

PR #784, opened 2025-02-07
Address exception classes for user vs. program errors

### LINKED ISSUE #782: Missing OSPModelDescription.xml causes logic_error to be thrown
If a user tries to use OSP-specific features like variable group connections in their `OspSystemStructure.xml`, but the required `*_OspModelDescription.xml` files are missing or not found, a `std::logic_error` will be thrown during parsing of the system structure file.

In other words, a user (runtime) error leads to an exception which signals a program (logic) error. Exceptions that represent runtime errors should always derive from `std::runtime_error`.

The source of the exception is `get_emd()` in `src/cosim/osp_config_parser.cpp`, which [throws](https://github.com/open-simulation-platform/libcosim/blob/bfd70935fa635eb77b6cd863c98f97a37b5749d4/src/cosim/osp_config_parser.cpp#L716) `std::out_of_range`, a subclass of `std::logic_error`. I'd say this is a slight misuse of this exception class.

Example of someone encountering this in the wild: open-simulation-platform/cosim-cli#115

restenb:
This PR fixes #782 by changing to `std::runtime_error` where the cause is incorrect configuration in the XMLs. 

While that squashes one loose chicken, a cursory glance at the codebase reveals several other places that could use similar changes. For instance, `std::invalid_argument` exceptions still abound in `scenario_parser.cpp`. And there probably are other places where `std::out_of_range` or `std::logic_error` may be dubiously applied.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,5 +9,7 @@ cmake-build-*/
 
 # Visual Studio Code
 .vscode/
+.clj-kondo/
+.lsp/
 
 CMakeUserPresets.json
--- a/src/cosim/osp_config_parser.cpp
+++ b/src/cosim/osp_config_parser.cpp
@@ -591,7 +591,7 @@ bool osp_config_parser::parse_boolean_value(const std::string& s)
         iss.clear();
         iss >> std::boolalpha >> bool_value;
         if (iss.fail()) {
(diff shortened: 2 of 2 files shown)


kyllingstad (approved):
LGTM

PR #785, opened 2025-02-24
State saving and serialization

### LINKED ISSUE #757: Enable saving the entire simulation state to disk and restoring it again

### LINKED ISSUE #779: Implement CBOR based de-/serializer for state import/export

davidhjp01:
Closes #757, closes #779

State saving-in-memory and serialization-to-disk support using  [cbor](https://cbor.io/).


restenb:
This looks great, I only have a few questions around usage. 

* Do we need to think about bounding the memory usage of FMUs where state-saving is used, e.g using a rolling buffer for in-memory state? Technically it's possible to call `save_state` on every simulation step, which will quickly increase memory usage. 
* On that note, we need to make sure there are no leaks and that it is clear when saved state can / should be freed. Even a small memory leak could lead to big problems if the state is saved thousands of times. It would make sense to test this with memory profiling a few times. 
* Should we document a "testing procedure" for ensuring that state save / restore is working as expected for user FMUs that are more complicated than `dahlquist`? You'd probably want to progress the simulation until `end_time` without saving state first, then comparing to runs where state was saved & restored at certain intervals.

kyllingstad:
> This looks great, I only have a few questions around usage.
> 
>  * Are there any implications to think about wrt. FMU solver type and restoring of intermediate state - thinking here of certain Runge-Kutta methods that sometimes take additional half-steps etc?

I don't think so, because any FMU with such issues would not be FMI compliant, and there is little we can do about that.

>  * Do we need to think about bounding the memory usage of FMUs where state-saving is used, e.g using a rolling buffer for in-memory state?

I don't think the FMI supports this. FMUs that have the capability flag `canNotUseMemoryManagementFunctions = false` receive pointers to `alloc()` and `free()` functions from Libcosim upon initialisation, and we could in principle implement those such that they draw from a limited, per-FMU memory pool. But there is no guarantee that the FMU actually uses them. Since an FMU can contain arbitrary machine code, it can allocate memory however it wants, and there is little we can do about it (except to run the subsimulators in separate, sandboxed, memory-limited processes). Secondly, those memory management functions are supposed be used to manage *all* of an FMU's memory, not just the memory it uses to save its state.

>Technically it's possible to call `save_state` on every simulation step, which will quickly increase memory usage.

As I see it, there are two main use cases for the save-restore-state functionality:

1. **Saving state persistently to resume a simulation from a known point later.** This is typically initiated by a *user* and should thus happen rarely, typically just once during a simulation run.
2. **Saving state temporarily as part of a rollback-based error estimation/control algorithm.** If any such algorithm saves state frequently, it is presumably because it needs to do so, and if it fails to release "old" states that are no longer needed, it is a bug – a memory leak, basically – in the algorithm.

>  * On that note, we need to make sure there are no leaks and that it is clear when saved state can / should be freed. Even a small memory leak could lead to big problems if the state is saved and restored thousands of times. It would make sense to test this with memory profiling a few times at least.

Agreed, we should make sure that Libcosim actually releases all internal resources associated with a saved state. But it is also here difficult/impossible to control what the FMUs do.

> * Should we document a "testing procedure" for ensuring that state save / restore is working as expected for user FMUs that are more complicated than `dahlquist`? You'd probably want to progress the simulation until `end_time` without saving state first, then comparing to runs where state was saved & restored at certain intervals. The results should be a close match though obviously they won't be exact.

Assuming that other settings like modifiers etc. are the same in the two cases, I actually think the results *should* be exactly equal. And if they aren't, it means that there is a bug in either Libcosim or the FMUs.

PR #789, opened 2025-05-14
Using proxyfmu/0.4.1 that uses increased frame size

davidhjp01:
Using proxyfmu 0.4.1 that is using increased frame size


.github/workflows/ci-conan.yml
```diff
@@ -50,6 +51,7 @@ jobs:
             --options="${{ matrix.option_shared }}" \
             --update \
             --build=missing \
+            --build=b2/* \
```
davidhjp01, line 54: b2 from conancenter uses "modern" GLIBC and fails linking on our build. Rebuilding it until b2 supports `--build=compatible:[pattern]`.

kyllingstad (approved):
:+1:

davidhjp01:
Need to fix cmake action that builds libcbor separately

PR #797, opened 2025-10-21
Don't crash if FMUs log with category=NULL

kyllingstad:
This fixes #796.

We don't have any straightforward way to test this right now, since our test FMUs aren't easily modifiable. Perhaps @SSkjong, who discovered the issue, can test with the FMU that triggered it?

### DIFF
--- a/src/cosim/fmi/v1/fmu.cpp
+++ b/src/cosim/fmi/v1/fmu.cpp
@@ -236,7 +236,8 @@ void log_message(
             break;
     }
     BOOST_LOG_SEV(log::logger(), logLevel)
-        << "[FMI status=" << statusName << ", category=" << category << "] "
+        << "[FMI status=" << statusName
+        << ", category=" << (category ? category : "") << "] "
         << msgBuffer.data();
 
     g_logMutex.lock();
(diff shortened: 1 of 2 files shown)


SSkjong:
Ser bra ut dette.

kyllingstad:
I don't understand the build failures. I don't think they are related to this particular PR, and I can't reproduce them locally even using the same Conan version as that used in CI.

kyllingstad:
@davidhjp01, @restenb, do any of you have any idea what is going on in the Linux CI builds here?

davidhjp01:
> @davidhjp01, @restenb, do any of you have any idea what is going on in the Linux CI builds here?

I also had the same issue, try this changes:
https://github.com/open-simulation-platform/libcosim/commit/62cd2ecc3c75299409f3dd4acc3a80300c333fc4

kyllingstad:
I think the fix in #801 should take care of the build issues here. I'll update as soon as that one is merged.

kyllingstad:
@davidhjp01, can we do a release after merging this? Other software depends on this fix.

davidhjp01 (approved):

PR #802, opened 2026-05-15
Upgrade CMake to v4

davidhjp01:
Upgrade CMake to v4

### DIFF
--- a/.github/workflows/ci-conan.yml
+++ b/.github/workflows/ci-conan.yml
@@ -28,7 +28,7 @@ jobs:
         run: |
           apt-get update
           apt-get install -y pipx cmake doxygen
-          pipx install conan
+          pipx install conan cmake
           echo "$HOME/.local/bin" >> "$GITHUB_PATH"
       - name: Configure Conan
         run: |
--- a/.github/workflows/ci-conan.yml
+++ b/.github/workflows/ci-conan.yml
@@ -39,6 +39,7 @@ jobs:
         run: |
           REFNAME="${GITHUB_REF#refs/*/}"
           VERSION="v$(<version.txt)"
+          cmake --version
           if [[ $GITHUB_REF == refs/tags/* ]] && [[ $REFNAME == $VERSION ]]; then CHANNEL="stable"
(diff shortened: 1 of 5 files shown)


conanfile.py
```diff
@@ -40,17 +40,17 @@ def set_version(self):
...
-        self.tool_requires("cmake/[>=3.19]")
+        self.tool_requires("cmake/[>=4.0]")
         self.requires("fmilibrary/[~2.3]")
         self.requires("libcbor/0.11.0")
         self.requires("libzip/[~1.11]")
         self.requires("ms-gsl/[>=3 <5]", transitive_headers=True)
         self.requires("boost/[~1.85]", transitive_headers=True, transitive_libs=True)  # Required by Thrift
         if self.options.proxyfmu:
-            self.requires("proxyfmu/0.4.1@osp/stable",
+            self.requires("proxyfmu/0.4.1@osp/testing-cmake-v4",
```
kyllingstad: I guess this should be reverted before merging.

.github/workflows/ci-conan.yml
```diff
@@ -28,7 +28,7 @@ jobs:
         run: |
           apt-get update
           apt-get install -y pipx cmake doxygen
-          pipx install conan
+          pipx install conan cmake
```
kyllingstad: Is this necessary? It seems strange to install CMake via the Python package system. Won't CMake be installed by Conan due to the `tool_requires()` directive in `conanfile.py`?

davidhjp01: Yeah, I thought so too, but since it was originally installing conan via `apt`, I just moved it here for updating it to the latest version. I can try removing it

kyllingstad: Maybe also try removing the APT install, so we are sure it's using the Conan-installed one.

conanfile.py
davidhjp01: I will updated it to stable version after releasing proxyfmu

kyllingstad (approved):

PR #803, opened 2026-07-09
Fix three ECCO power-bond bugs and add regression tests

joakimono:
# Fix three ECCO power-bond bugs

## Background

Three bugs were present in the ECCO algorithm's power-bond support.
Two caused crashes or spurious errors at runtime; one caused a silent
double-registration in a test that happened to pass only by accident.

---

## Bug 1 — `add_power_bond()` never exposes bond variables for getting

**File:** `src/cosim/algorithm/ecco_algorithm.cpp`

### What went wrong

`ecco_algorithm::impl::add_power_bond()` stores four `variable_id` values
(the two force/velocity pairs that make up the bond) but never calls
`expose_for_getting()` on any of them.

`adjust_step_size()` later calls `sim->get_real(ref)` on all four variables
every step. `get_real()` reads from the simulator's get-cache, which only
contains variables that were previously registered via `expose_for_getting()`.
For FMU **output** variables that also appear in a `connect_variables()`
connection, this registration happened to occur as a side-effect. For FMU
**input** variables (`chassis.force`, `wheel.in_vel`) it never did, because
`connect_variables()` registers inputs only for *setting*, not for getting.

The crash was therefore latent for the two bond-input variables:

```
terminate: Variable with reference N not found in exposed variables
```

### Why it was masked

The default `file_observer` (constructed without a config) calls
`initialize_default()` which exposes **all** non-local variables for getting,
accidentally making the bond inputs readable. Any test or application using
the unconfigured observer never saw the crash.

### How it is triggered

A `file_observer` constructed from a `LogConfig.xml` that omits the bond
input variables (`force`, `in_vel`) exposes only the listed variables. When
`adjust_step_size()` then tries to read the unlisted inputs, the get-cache
lookup throws.

### Fix

Call `expose_for_getting()` for all four bond variables at the top of
`add_power_bond()`, before storing them. The call is idempotent — if an
observer already exposed the variable, the second registration is a no-op.

```cpp
// ecco_algorithm.cpp — add_power_bond()
simulators_.at(input_a.simulator).sim->expose_for_getting(input_a.type, input_a.reference);
simulators_.at(output_a.simulator).sim->expose_for_getting(output_a.type, output_a.reference);
simulators_.at(input_b.simulator).sim->expose_for_getting(input_b.type, input_b.reference);
simulators_.at(output_b.simulator).sim->expose_for_getting(output_b.type, output_b.reference);
```

### Regression test

`tests/ecco_algorithm_powerbond_logconfig_test.cpp` — loads the quarter-truck
via `inject_system_structure`, then attaches a `file_observer_config` that
deliberately omits `chassis.force` and `wheel.in_vel` from the logged variable
list, and runs a short simulation. Before the fix this test crashes; after the
fix it completes cleanly.

---

## Bug 2 — `add_power_bonds()` validation check fires inside the per-bond loop

**File:** `src/cosim/osp_config_parser.cpp`

### What went wrong

After each call to `systemStructure.add_power_bond()`, the function checked:

```cpp
if (uniqueCount != numPowerBonds) { throw ...; }
```

`uniqueCount` was computed up front from **all** connection entries across
**all** bonds, so with two bonds it always equalled 2. But `numPowerBonds`
was queried from `systemStructure` inside the loop, so after the first
iteration it was only 1. The check `2 != 1` fired immediately, making any
system with two or more power bonds impossible to load.

### Fix

Move the sort + uniqueCount + validation block outside the loop so it runs
once, after all bonds have been registered:

```cpp
    for (const auto& pbName : uniquePowerBondNames) {
        ...
        systemStructure.add_power_bond(pbName, powerbond);
    }                                           // ← loop ends here

    // validation runs once, with all bonds present
    std::sort(powerBondNames.begin(), powerBondNames.end());
    auto uniqueCount = ...;
    auto numPowerBonds = static_cast<int>(systemStructure.get_power_bonds().size());
    if (uniqueCount != numPowerBonds) { throw ...; }
```

### Regression test

`tests/ecco_algorithm_two_bond_test.cpp` — loads
`OspSystemStructure_MultiBond.xml`, which defines two independent
chassis+wheel pairs (`chassis1`/`wheel1` and `chassis2`/`wheel2`) each
connected by a separate power bond (`bond1`, `bond2`). Before the fix,
`load_osp_config()` throws on this file; after the fix the config loads and
the simulation runs for 0.1 s without error.

`tests/data/fmi2/quarter_truck/OspSystemStructure_MultiBond.xml` is the new
test data file. Each pair is physically independent; the same Chassis.fmu and
Wheel.fmu are used for both, so no new FMUs are required.

---

## Bug 3 — `ecco_algorithm_multi_bond_test` uses wrong variable refs and double-registers the bond

**File:** `tests/ecco_algorithm_multi_bond_test.cpp`

### What went wrong

Two mistakes in the test:

1. **Duplicate bond registration.** `inject_system_structure()` already parses
   the power bond from `OspSystemStructure.xml` and calls `add_power_bond()`
   internally. The test then called `add_power_bond()` a second time manually,
   producing two energy-residual accounting entries for the same bond and
   doubling the power-residual calculation.

2. **Wrong hardcoded value references.** The manually constructed `variable_id`
   objects used refs 19, 22, 26, 24. Inspection of the FMU model descriptions
   shows these are ECCO step-size tuning parameters
   (`stepSizeCtrlSafetyFactor`, `timeStepSlowStart`, etc.), not the bond
   variables. The correct refs are:

   | Variable | FMU causality | Value reference |
   |---|---|---|
   | `chassis.velocity` | output | 23 |
   | `chassis.force` | input | 4 |
   | `wheel.out_spring_damper_f` | output | 15 |
   | `wheel.in_vel` | input | 7 |

   The test passed despite the wrong refs only because the default
   `csv_observer` exposed all variables for getting, masking Bug 1 for
   the bond-input refs, and the wrong parameter refs happened not to crash
   the simulation.

### Fix

Remove the duplicate `add_power_bond()` call. Correct the `variable_id`
refs used for the `time_series_observer` observations. Remove the two dead
output vectors (`wheelGVels`, `groundVels`) that were allocated but never
filled.

### DIFF
--- a/src/cosim/algorithm/ecco_algorithm.cpp
+++ b/src/cosim/algorithm/ecco_algorithm.cpp
@@ -309,6 +309,10 @@ public:
 
     void add_power_bond(cosim::variable_id input_a, cosim::variable_id output_a, cosim::variable_id input_b, cosim::variable_id output_b)
     {
+        simulators_.at(input_a.simulator).sim->expose_for_getting(input_a.type, input_a.reference);
+        simulators_.at(output_a.simulator).sim->expose_for_getting(output_a.type, output_a.reference);
+        simulators_.at(input_b.simulator).sim->expose_for_getting(input_b.type, input_b.reference);
+        simulators_.at(output_b.simulator).sim->expose_for_getting(output_b.type, output_b.reference);
         energies_.emplace_back();
         energies_.emplace_back();
         inputVariables_.push_back(input_a);
--- a/src/cosim/osp_config_parser.cpp
+++ b/src/cosim/osp_config_parser.cpp
@@ -891,17 +891,20 @@ void add_power_bonds(const std::vector<osp_config_parser::PowerBondConnection>& 
         }
 
         systemStructure.add_power_bond(pbName, powerbond);
+    }
 
-        // Check that the number of unique power bond names is equal to the number of power bonds. Otherwise, it is not possible to correctly connect the bonds.
-        std::sort(powerBondNames.begin(), powerBondNames.end());
-        auto uniqueCount = static_cast<int>(std::unique(powerBondNames.begin(), powerBondNames.end()) - powerBondNames.begin());
-        auto numPowerBonds = static_cast<int>(systemStructure.get_power_bonds().size());
+    // Check that the number of unique power bond names is equal to the number of
+    // power bonds. This must be verified after all bonds have been added; checking
+    // inside the loop would compare a partial count against the full unique-name
+    // count and spuriously throw when more than one bond is configured.
+    std::sort(powerBondNames.begin(), powerBondNames.end());
+    auto uniqueCount = static_cast<int>(std::unique(powerBondNames.begin(), powerBondNames.end()) - powerBondNames.begin());
+    auto numPowerBonds = static_cast<int>(systemStructure.get_power_bonds().size());
 
-        if (uniqueCount != numPowerBonds) {
-            std::ostringstream oss;
-            oss << "The number of powerbonds (" << numPowerBonds << ") is not equal to the number of unique power bond names (" << uniqueCount << ") found in the configured system. Power bond names must be unique pr. bond, that is found on only and exactly the two VariableConnections that form the bond.";
-            throw std::runtime_error(oss.str());
-        }
+    if (uniqueCount != numPowerBonds) {
+        std::ostringstream oss;
+        oss << "The number of powerbonds (" << numPowerBonds << ") is not equal to the number of unique power bond names (" << uniqueCount << ") found in the configured system. Power bond names must be unique pr. bond, that is found on only and exactly the two VariableConnections that form the bond.";
+        throw std::runtime_error(oss.str());
     }
 }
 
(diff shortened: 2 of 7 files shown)


kyllingstad (approved):
LGTM, thanks!

joakimono:
> LGTM, thanks!

Should I add commits on this branch in an attempt to fix CI issues?

kyllingstad:
> Should I add commits on this branch in an attempt to fix CI issues?

I just re-ran the workflow on `master`, and it looks like our whole CI is broken (again). It's probably cleaner to fix it all in a separate PR.

restenb (approved):
Thanks, some good and necessary catches here! 

I tasked our new assistant with fixing the CI issues, which it did after some attempts (notably by changing windows build to 2022 version because of issues with fmilibrary):
https://github.com/open-simulation-platform/libcosim/pull/804

(oh, and based on the ... verbosity of the PR description, perhaps you have also met our new assistant?) 😸 

In order to support windows 25 builds again we'd need to update the fmilibrary build and/or conan package to work with newer cmake.

PR #805, opened 2026-08-18
Usability improvements for power bonds and ECCO

restenb:
I wanted to improve the usability of the current implementation a bit, as it had some notable warts. This PR therefore makes the following changes:

1. Removed the `causality` attribute from the `<Variable>` element in the XML. This was always unnecessary as the causality can be read directly from the model description which has already been parsed. 
2. Made it so that the individual power bonds in an `ecco_algorithm` are addressable by the configured name of the bond to avoid having to deal with opaque indices. 
3. Added convenience structs `power_bond_info `and `power_bond_state ` for collecting the information (the configured bond name and it's associated `variable_id`s), and the runtime state associated with each power bond. There is also more state now added to these structs, in particular accumulated energy residual pr. bond, and the per-bond error contribution to the error estimate used to adjust the time steps. This provides necessary visibility into which bond(s) are contributing the most to the overall error. More state from the algorithm can easily be exposed via the same mechanism. 
4. Added API functions that interact with these state objects, rather than indices or vectors of energies directly. 
5. Avoided accumulating the energies inside of the algorithm, which would have grown unbounded with the simulation time. 
6. Added a new `ecco_observer ` instead for observing power bonds and their states, and associated test. The observer also supports a ring buffer with configurable size (like the `time_series_observer`), and simple energy statistics (max/min/mean/RMS), useful for tracking how well the algorithm performs over an entire run. 

Note: As there are some minor breaking changes with new tests added in #803, I suggest merging that first and I will update the tests.


data/xsd/OspSystemStructure.xsd
```diff
@@ -78,7 +78,6 @@
     <xs:complexType name="variableEndpoint">
         <xs:attribute name="simulator" use="required" type="xs:string"/>
         <xs:attribute name="name" use="required" type="xs:string"/>
-        <xs:attribute name="causality" type="xs:string"/>
```
davidhjp01, line 81: Should we postpone this removal and generate a warning message that this will be removed in the upcoming release, if used? Because I think this might break some apps.

tests/ecco_observer_test.cpp
davidhjp01, line 1: Should we check if the state can also be exported into a file and loadable?

data/xsd/OspSystemStructure.xsd
restenb: As I recall this was added when power bonds was first released, which was last year? I'd be a bit surprised if it broke any apps except our own. Would rather just remove it now tbh ...

tests/ecco_observer_test.cpp
restenb: I didn't really consider state saving now, in order for that to actually work we'd have to at least serialize all the relevant ecco algorithm state including the per-bond energies. For safety we should probably also re-validate that the bonds actually still exist in the imported state. 

Suggest postponing this to a separate PR when there is need to use ECCO with state saving.

data/xsd/OspSystemStructure.xsd
davidhjp01: Aha, I thought it was used for general connection, then nvm

PR #807, opened 2026-09-18
Fmi2 directional derivatives

davidhjp01:
# Add FMI 2 directional derivative support

## Summary

Adds validated FMI 2.0 directional-derivative support to
`fmi::v2::slave_instance`, lifecycle-aware FMU state handling, deterministic
fixture coverage, and the dependency/CI updates required to build and test it.

## Changes

- Add `provides_directional_derivatives()` and
  `get_directional_derivative()`.
- Read and cache directional-derivative metadata from
  `modelDescription.xml`.
- Validate lifecycle, value-reference type, reference validity, and seed and
  sensitivity sizes before calling the FMU.
- Compute the Jacobian-vector product:

  ```text
  sensitivity = d(unknowns) / d(knowns) * seed
  ```

- Accept FMI `OK` and `Warning`; surface `Discard`, `Error`, `Fatal`, and
  `Pending` as errors.
- Preserve caller output when the FMU call fails.
- Track `instantiated`, `initialization`, `step`, and `terminated` states.
- Serialize lifecycle state in export scheme 1 while retaining safe scheme 0
  compatibility.
- Add cleanup and ownership checks for imported FMU state.

## Test coverage

- Add a deterministic FMI 2 Co-Simulation fixture with three Real inputs, two
  Real outputs, a fixed 2-by-3 Jacobian, configurable statuses, a call counter,
  and a non-Real reference.
- Extend `fmi_v2_fmu_unittest` for lifecycle behavior, expected derivatives,
  reordered references, invalid inputs, status handling, output preservation,
  and unsupported capabilities.
- Build the pinned Reference-FMUs `v0.0.41` and package the fixture as
  `Fmi2Derivatives.fmu`; expose them through `REFERENCE_FMU_V2_DIR`.

## Dependency and CI updates

- Upgrade FMI Library to `~3.0.4` and proxyFMU to `0.4.3@osp/testing`.


.github/workflows/ci-cmake.yml
```diff
@@ -15,45 +17,70 @@ jobs:
...
+            libboost-all-dev \
+            libmsgsl-dev \
+            libyaml-cpp-dev \
+            libxerces-c-dev \
+            libzip-dev \
+            wget \
+            zipcmp \
+            zipmerge \
+            ziptool
+          rm -rf /var/lib/apt/lists/*
```
kyllingstad: Is this necessary?

src/cosim/fmi/v2/fmu.cpp
```diff
@@ -698,7 +714,59 @@ slave::state_index slave_instance::import_state(
...
+        // Validate wrapper state before asking the FMU to allocate a state.
+        savedState.setupComplete = exportedState.get<bool>("setup_complete");
+        savedState.simStarted = exportedState.get<bool>("simulation_started");
+        if (schemeVersion == 0) {
+            if (savedState.setupComplete && !savedState.simStarted) {
+                throw error(
+                    make_error_code(errc::bad_file),
+                    "The serialized state of subsimulator '" + instanceName_
+                        + "' uses scheme 0 lifecycle data that is ambiguous "
+                          "between Initialization Mode and Terminated");
```
kyllingstad: I think I understand why this is done, namely to

1. support importing of old (scheme version 0) state data when this is possible without ambiguity
2. fail when old data cannot be loaded without ambiguity

However, I think it is introducing unnecessary complexity. Firstly, the error message will not make much sense to users, and it's not an issue they can do anything about. Secondly, I strongly suspect that very few people are actively using this feature yet, and so maintaining compatibility with previous schemes is not really something we need to prioritise at this point. (It becomes much more important once libcosim passes the 1.0 version mark.)

So to keep the code simple as long as we can, I would just keep the earlier test (at old/new line no. 680/696) as it was, namely throw if `schemeVersion != export_scheme_version`, and drop any further version-conditional branching.

Unless I'm missing something, I think we can even entirely get rid of the old "setup complete" and "simulation started" lifecycle stage flags, both as class member variables and exported state fields. The new enum subsumes both, right?

src/cosim/fmi/v2/fmu.cpp
```diff
@@ -742,6 +808,125 @@ fmi2_import_t* slave_instance::fmilib_handle() const
...
+            make_error_code(errc::unsupported_feature),
+            instanceName_ + ": FMU does not provide directional derivatives");
+    }
+    if (lifecycleState_ != lifecycle_state::initialization &&
+        lifecycleState_ != lifecycle_state::step &&
+        lifecycleState_ != lifecycle_state::terminated) {
+        throw error(
+            make_error_code(errc::model_error),
+            instanceName_ +
+                ": invalid lifecycle operation: GetDirectionalDerivative");
```
kyllingstad: I think this error message is misleading. Getting directional derivatives is not a "lifecycle operation", it is an operation that may be invalid at the current state in the FMU's lifecycle. I'm also not sure whether `errc::model_error` is the right error code for this. Maybe we need a new one, e.g. `errc::invalid_operation` or something. (It wouldn't be unique to this particular situation; there are many cases in the FMI spec where certain operations are prohibited at a certain lifecycle stage. Could also be used for the "invalid (un)known" errors further down in this function.)

src/cosim/fmi/v2/fmu.cpp
```diff
@@ -742,6 +808,125 @@ fmi2_import_t* slave_instance::fmilib_handle() const
...
+        const bool validUnknown = initializing
+            ? contains(initialUnknownReferences_, reference)
+            : eligibleOutput ||
+                contains(derivativeReferences_, reference);
+        if (variable == nullptr || !validUnknown) {
+            throw error(
+                make_error_code(errc::model_error),
+                instanceName_ + ": directional derivative unknown value reference " +
+                    std::to_string(reference) +
+                    " is not a mode-appropriate Real ModelStructure entry");
```
kyllingstad: I worry that this error message will not be very informative to users. *I* don't understand it, at least. (Also a candidate for a new `errc::invalid_operation` code, by the way.)

src/cosim/fmi/v2/fmu.cpp
```diff
@@ -742,6 +808,125 @@ fmi2_import_t* slave_instance::fmilib_handle() const
...
+    COSIM_INPUT_CHECK(seed.size() == knowns.size());
+    COSIM_INPUT_CHECK(sensitivity.size() == unknowns.size());
+
+    const auto contains = [](const auto& references, value_reference reference) {
+        return std::find(
+                   references.begin(),
+                   references.end(),
+                   reference) != references.end();
+    };
+    const bool initializing =
```
kyllingstad: It seems unnecessary to introduce a new variable for this. Just use the actual comparison expression when it's needed.

tests/fmi2_derivative_fixture/model.c
```diff
@@ -0,0 +1,758 @@
...
+    instance_of(component)->time = startTime;
+    update_outputs(instance_of(component));
+    return fmi2OK;
+}
+
+
+fmi2Status fmi2EnterInitializationMode(fmi2Component component)
+{
+    if (require_instance(component) != fmi2OK) {
+        return fmi2Fatal;
```
kyllingstad, line 223: Nitpicking here, but why would this be `fmi2Fatal` rather than just `fmi2Error`? The difference is supposed to be that an "error" means that this model instance is now invalid, while "fatal" means that *all* model instances are invalid. Same pattern repeats in several functions below.

tests/fmi2_derivative_fixture/model.c
```diff
@@ -0,0 +1,758 @@
...
+}
+
+
+fmi2Status fmi2GetString(
+    fmi2Component component,
+    const fmi2ValueReference references[],
+    size_t count,
+    fmi2String values[])
+{
+    (void)references;
```
kyllingstad, line 424: What is going on here? If it's just to avoid an "unused variable" error, wouldn't it be better to comment out the parameter names in the function declaration? More of this below.

tests/CMakeLists.txt
```diff
@@ -1,3 +1,114 @@
...
+        message(FATAL_ERROR
+            "Failed to clone Reference-FMUs: ${reference_fmus_clone_error}")
+    endif()
+endif()
+
+set(REFERENCE_FMUS_V2_BINARY_DIR
+    "${CMAKE_CURRENT_BINARY_DIR}/reference-fmus-v2-build")
+set(REFERENCE_FMUS_V2_STAMP_DIR
+    "${CMAKE_CURRENT_BINARY_DIR}/reference-fmus-v2-stamp-${REFERENCE_FMUS_VERSION}")
+ExternalProject_Add(
```
kyllingstad, line 41: I guess building the whole reference FMUs collection will add substantially to our build times? What is the rationale for doing so, is it strictly needed? (If we do keep this, we should be consistent and also take the _Dahlquist_ FMU from that repo, which we currently use in at least one test and keep in binary form under [`tests/data/fmi2`](https://github.com/open-simulation-platform/libcosim/tree/master/tests/data/fmi2).)

kyllingstad (changes requested):
Thanks! Great to finally have some more FMI 2.0 support… :D I have a few comments, mostly on form rather than substance.

.github/workflows/ci-cmake.yml
davidhjp01: Actually there are some cmake cmds added in the upcoming PR that uses these tools for updating/generating FMUs. But I can remove these and add them in the later PR
Or do you mean `rm`? (I noticed these zip tools were already there previously)

kyllingstad: I meant the `rm`, yes. Looks like it's clearing APT's package metadata cache for no discernible reason.

tests/fmi2_derivative_fixture/model.c
davidhjp01: It is just when component is NULL, checked by require_instance, the instance is unrecoverable so it is `fmi2Fatal` instead of `fmi2Error`.

src/cosim/fmi/v2/fmu.cpp
davidhjp01: Updated to throw invalid_operation and text

src/cosim/fmi/v2/fmu.cpp
davidhjp01: Removed setup complete and simulation started boolean vars, and using lifecycle enum directly

tests/fmi2_derivative_fixture/model.c
davidhjp01: Seems MSVC rejects commented out parameter names: https://learn.microsoft.com/en-us/cpp/error-messages/compiler-errors-1/compiler-error-c2055?view=msvc-170. We can add ignore for this warning in cmake, what do you think?

tests/CMakeLists.txt
davidhjp01: In the upcoming PRs, I made many fmi3 test-cases using reference fmus, also used its framework to build custom FMUs for testing new features (like in this PR). This is to avoid committing binaries as the upstream changes the FMU logic. 

I think we can disable building tests by default, if we want to avoid increase in build time. I will update the tests to use Dahlquist fmu from this repo.

src/cosim/fmi/v2/fmu.cpp
davidhjp01: Updated

tests/fmi2_derivative_fixture/model.c
restenb: Are there any pre-made test models for this (and upcoming FMI3 features) available directly from FMI that we could use instead of specifying our own?

tests/CMakeLists.txt
restenb: Our established pattern is to provide the FMU files directly, and not rely on an external network dependency at configure time. I suspect the stock FMI3 test FMUs don't change so often that this automation is all that necessary either. 

At the very least the clone should be kept off the regular build path and gated by default, e.g. `LIBCOSIM_TEST_DIRECTIONAL_DERIVATIVES` with default `OFF`.

src/cosim/fmi/v2/fmu.cpp
restenb: > Unless I'm missing something, I think we can even entirely get rid of the old "setup complete" and "simulation started" lifecycle stage flags, both as class member variables and exported state fields. The new enum subsumes both, right?

Agree. It is simpler to collapse these states into the `lifecycleState_` enum as you point out, and as David has perhaps just done. :) They were just used to distinguish between those states anyway.

tests/fmi2_derivative_fixture/model.c
kyllingstad: I guess this is fine. The `fmi2Fatal` error code is quite poorly defined in the spec, in my opinion. The FMI 2 spec says that it means that “*all* model instances are irreparably corrupted”, i.e., not just this one instance. But then the spec goes on to give examples like access violation and integer divide by zero, both of which lead to undefined behaviour, not a trustworthy error code.

tests/fmi2_derivative_fixture/model.c
kyllingstad: No, then I prefer to keep it as-is.

include/cosim/fmi/v2/fmu.hpp
```diff
@@ -186,21 +186,46 @@ class slave_instance : public fmi::slave_instance
...
     fmi2_import_t* fmilib_handle() const;
 
+    /// Reports whether the FMU provides directional derivatives.
+    bool provides_directional_derivatives() const noexcept;
+
+    /**
+     *  Computes directional derivatives of unknown Real variables with respect
+     *  to known Real variables.
+     */
+    void get_directional_derivative(
```
restenb: Any thought to how `algorithm` should be able to access this? IIRC, these functions are masked in the `slave_simulator` interface, so we are missing something like `simulator->get_directional_derivative(..)` to actually make use of this higher up in the food chain. 

It's of course OK if this is done in a future PR since we don't have anything that needs to consume directional derivatives right here and now.

src/cosim/fmi/v2/fmu.cpp
```diff
@@ -631,7 +625,7 @@ void slave_instance::release_state(state_index state)
 // state that needs to be serialized may change, we need some versioning.
 // Increment this number whenever the "exported state" changes form, and
 // always consider whether backwards compatibility measures are warranted.
-constexpr std::int32_t export_scheme_version = 1;
+constexpr std::int32_t export_scheme_version = 0;
```
kyllingstad: This should not be reverted, IMO. It *is* a new scheme version, my point was just that shouldn't worry about maintaining compatibility with the previous one.

kyllingstad (changes requested):
Almost there now. :)

src/cosim/fmi/v2/fmu.cpp
davidhjp01: Made it back to 1

kyllingstad (approved):
Nice, thanks for doing this!

PR #808, opened 2026-09-21
FMI3 Types definitions

davidhjp01:
## Summary

Extends the shared model-description layer with FMI 3 value types, arrays,
metadata, and detailed step outcomes while preserving existing FMI 1/2
behavior.

## Changes

- Add FMI 3 types: Float32, signed and unsigned integer widths, Binary, and
  Clock.
- Add FMI 3 causality values, typed value references, Enumeration and Binary
  values, and type-safe scalar/array storage.
- Add `variable_value` with:
  - exact type and size access;
  - scalar and flat row-major array values;
  - dimension validation;
  - overflow-safe `flat_size()` calculation.
- Extend variable and model metadata with:
  - FMI 3 instantiation tokens;
  - declared types, units, display units, bounds, and nominal values;
  - aliases, derivatives, previous variables, clocks, and array dimensions;
  - Enumeration, Binary, and Clock definitions.
- Add `fixed_shape()` to distinguish fixed arrays from arrays sized by
  structural parameters.
- Add `step_result_info`/`step_outcome` for actual completion time, early
  return, event handling, termination requests, and retryability.
- Make existing execution paths exhaustive:
  - unsupported initial-value types now fail explicitly;
  - generic value validation recognizes the new typed values.

## Compatibility and scope

- Existing FMI 1/2 scalar types and execution behavior remain supported.
- New metadata and value types provide the common foundation for FMI 3
  importer and typed-access work.
- This branch does not by itself add FMI 3 FMU loading or runtime get/set
  operations.


kyllingstad:
I'll review this one once #807 is done, in case there are changes which propagate to this PR.

include/cosim/fmi/v2/fmu.hpp
```diff
@@ -186,21 +186,50 @@ class slave_instance : public fmi::slave_instance
...
     fmi2_import_t* handle_;
 
     bool setupComplete_ = false;
     bool simStarted_ = false;
+    lifecycle_state lifecycleState_ = lifecycle_state::instantiated;
+    bool providesDirectionalDerivatives_ = false;
+    std::vector<value_reference> outputReferences_;
+    std::vector<value_reference> derivativeReferences_;
+    std::vector<value_reference> initialUnknownReferences_;
+    std::vector<value_reference> continuousStateReferences_;
```
restenb: One thing that immediately confused me is that there is no `/fmi/v3` with the actual implementation of a `v3::fmu`. If the intention was to just add the new type vocabulary in this PR that's still fine. But right now this PR is a large new API surface that immediately panics if anybody tries to use it.

include/cosim/model_description.hpp
```diff
@@ -30,39 +33,81 @@ using value_reference = std::uint32_t;
 /// Variable data types.
 enum class variable_type
```
restenb: We can at least see that we are going from ~5 to ~15 types in this enum. This is used all over the code base in switch statements like this one from `fixed_step_algorithm.cpp`:

Image

The switch only exists to pick the correct method to call e.g. `target->set_real(ref, source->get_real(ref))`. Maybe we can hide these details behind a type agnostic function or interface so we can instead just write something like `target->set(c.target, source->get(c.source))` and do away with all the duplicate switch cases by hiding it inside that boundary. But we still want to _know_ the type outside that interface, so the mechanism has to be thought through a bit.

